PR #16, opened 2022-11-30
Support batching

### LINKED ISSUE #15: Batch contributions for the same reporting origin together
The report scheme permits multiple contributions in the same report. We should consolidate some histogram contributions to reduce the volume of reports. For example, calls to sendHistogramReport() during the same Shared Storage operation could be batched together.

alexmturner:
Resolves #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ Author: Alex Turner (alexmt@chromium.org)
     - [Enabling](#enabling)
     - [Debug keys](#debug-keys)
     - [Duplicate debug report](#duplicate-debug-report)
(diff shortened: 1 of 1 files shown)


alexmturner:
@johnivel @csharrison could you PTAL (in particular at the suggested batching scopes)? Thanks!

johnivdel (approved):
LGTM

johnivdel:
It might worth pointing out something about the size of reports leaking information, which necessitates padding / etc

alexmturner:
Added a comment

PR #27, opened 2023-04-14
Spec: replace global scope names

alexmturner:
Updates how global scopes are referred to in the spec. Requires some minor changes to the Protected Audiences (formerly FLEDGE) spec for this change to make sense. Those are being made [here](https://github.com/WICG/turtledove/pull/546).

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/27.html" title="Last updated on Apr 19, 2023, 6:50 PM UTC (e7ce93c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/27/5b8a817...e7ce93c.html" title="Last updated on Apr 19, 2023, 6:50 PM UTC (e7ce93c)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -94,7 +94,7 @@ Each item must be either a {{PAHistogramContribution}} or a
 {{PAExtendedHistogramContribution}}.
 
 Note: The steps to process the [=contributions cache=] are defined separately for
-each [=worklet type identifier=].
+each [=context type=].
 
 Issue: Do we need to spec enableDebugMode?
 
--- a/spec.bs
+++ b/spec.bs
@@ -18,7 +18,6 @@ Assume Explicit For: on
 <pre class="anchors">
 urlPrefix: https://wicg.github.io/turtledove/; type: interface
     text: InterestGroupBiddingScriptRunnerGlobalScope
-    text: InterestGroupScriptRunner
     text: InterestGroupScriptRunnerGlobalScope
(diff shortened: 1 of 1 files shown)


alexmturner:
@linnan-github, could you PTAL? Thanks! (I won't land it until https://github.com/WICG/turtledove/pull/546 does.)

spec.bs
```diff
@@ -91,7 +94,7 @@ Each item must be either a {{PAHistogramContribution}} or a
 {{PAExtendedHistogramContribution}}.
 
 Note: The steps to process the [=contributions cache=] are defined separately for
-each {{WorkletGlobalScope}}.
+each [=context type=].
```
linnan-github, line 97: Context type is fledge or shared storage, isn't this per global scope?

alexmturner: What I'm trying to convey here is that there's a single algorithm defined for all of FLEDGE. (Separately, I have to transition this spec to per-auction rather than per-global scope processing.) Lmk if you have any ideas to reword

linnan-github: Ah yes, I misread this as the definition. Sorry for the noise.

linnan-github:
LGTM!

alexmturner:
Thanks, Nan!

PR #28, opened 2023-04-18
Spec: Add report sending logic

alexmturner:
Heavily based off the Attribution Reporting API spec

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/28.html" title="Last updated on May 3, 2023, 9:37 PM UTC (3cc328d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/28/5ca6783...3cc328d.html" title="Last updated on May 3, 2023, 9:37 PM UTC (3cc328d)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -25,7 +25,6 @@ urlPrefix: https://wicg.github.io/shared-storage/; type: interface
     text: SharedStorageWorklet
     text: SharedStorageWorkletGlobalScope
 spec: hr-time; type: dfn; urlPrefix: https://w3c.github.io/hr-time/
-    text: current unsafe time; url: #wall-clock-unsafe-current-time
     text: current wall time; url: #dfn-current-wall-time
     text: duration; url: #dfn-duration
     text: duration from; url: #dfn-duration-from
@@ -135,7 +134,7 @@ Immediately after an operation completes, [=process the Shared Storage contribut
 
 Issue: How to handle batching different operation invocations properly. This doesn't work for simultaneous operations. Also disallow usage outside an operation.
 
-To <dfn>process the Shared Storage contributions cache</dfn> given a [=list=] |contributionsCache| and a {{SharedStorageWorkletGlobalScope}} |scope|, perform the [=PrivateAggregation/report creation and scheduling steps=] with |scope|'s [=relevant settings object=], "<code>[=context type/shared-storage=]</code>" and |contributionsCache|.
+To <dfn>process the Shared Storage contributions cache</dfn> given a [=list=] |contributionsCache| and a {{SharedStorageWorkletGlobalScope}} |scope|, perform the [=PrivateAggregation/report creation and scheduling steps=] with |scope|'s [=relevant settings object=]'s [=environment settings object/origin=], "<code>[=context type/shared-storage=]</code>" and |contributionsCache|.
 
 Exposing to FLEDGE {#fledge}
 ============================
@@ -186,7 +185,7 @@ To <dfn>process the FLEDGE contributions cache</dfn> given a [=list=] |contribut
 1. Let |filledInContributions| be a new [=list/is empty|empty=] [=list=].
 1. [=list/iterate|For each=] |contribution| of |contributionsCache|:
     1. [=list/Append=] the result of [=filling in the contribution=] given |contribution| to |filledInContributions|.
-1. Perform the [=PrivateAggregation/report creation and scheduling steps=] with |scope|'s [=relevant settings object=], "<code>[=context type/fledge=]</code>" and |filledInContributions|.
+1. Perform the [=PrivateAggregation/report creation and scheduling steps=] with |scope|'s [=relevant settings object=]'s [=environment settings object/origin=], "<code>[=context type/fledge=]</code>" and |filledInContributions|.
 
 
 Structures {#structures}
@@ -279,8 +278,7 @@ Issue: This would ideally be replaced by a more descriptive algorithm in Infra. 
 Scheduling reports {#scheduling-reports}
 ----------------------------------------
 
-To perform the <dfn algorithm for="PrivateAggregation">report creation and scheduling steps</dfn> with an [=environment settings object=] |context|, a [=context type=] |api| and a [=list=] of {{PAHistogramContribution}}s |contributions|:
-1. Let |reportingOrigin| be |context|'s [=environment settings object/origin=].
+To perform the <dfn algorithm for="PrivateAggregation">report creation and scheduling steps</dfn> with an [=origin=] |reportingOrigin|, a [=context type=] |api| and a [=list=] of {{PAHistogramContribution}}s |contributions|:
 1. [=Assert=]: |reportingOrigin| is a [=potentially trustworthy origin=].
 1. Let |truncatedContributions| be a new [=list/is empty|empty=] [=list=].
 1. If |contributions| has a [=list/size=] greater than [=maximum report contributions=]:
@@ -291,9 +289,9 @@ To perform the <dfn algorithm for="PrivateAggregation">report creation and sched
 1. [=set/iterate|For each=] |contribution| of |truncatedContributions|:
     1. [=Assert=]: |contribution|["|value|"] is non-negative.
     1. Add |contribution|["|value|"] to |contributionSum|.
-1. Let |sufficientBudget| be the result of [=PrivateAggregation/consuming budget if permitted=] given |contributionSum|, |reportingOrigin|, |api| and |context|'s [=current wall time=].
+1. Let |sufficientBudget| be the result of [=PrivateAggregation/consuming budget if permitted=] given |contributionSum|, |reportingOrigin|, |api| and the [=current wall time=].
 1. If |sufficientBudget| is false, return.
-1. Let |report| be the result of [=PrivateAggregation/obtaining an aggregatable report=] given |reportingOrigin|, |api|, |truncatedContributions| and |context|'s [=current wall time=].
+1. Let |report| be the result of [=PrivateAggregation/obtaining an aggregatable report=] given |reportingOrigin|, |api|, |truncatedContributions| and the [=current wall time=].
 1. [=set/Append=] |report| to the user agent's [=aggregatable report cache=].
 
 Issue: Do we need to ensure the reports aren't queued after being sent?
--- a/spec.bs
+++ b/spec.bs
@@ -290,7 +290,7 @@ To perform the <dfn algorithm for="PrivateAggregation">report creation and sched
 1. Let |contributionSum| be 0.
 1. [=set/iterate|For each=] |contribution| of |truncatedContributions|:
     1. [=Assert=]: |contribution|["|value|"] is non-negative.
-    1. Add |value| to |contributionSum|.
+    1. Add |contribution|["|value|"] to |contributionSum|.
 1. Let |sufficientBudget| be the result of [=PrivateAggregation/consuming budget if permitted=] given |contributionSum|, |reportingOrigin|, |api| and |context|'s [=current wall time=].
 1. If |sufficientBudget| is false, return.
 1. Let |report| be the result of [=PrivateAggregation/obtaining an aggregatable report=] given |reportingOrigin|, |api|, |truncatedContributions| and |context|'s [=current wall time=].
@@ -302,7 +302,7 @@ Issue: Do we need to address user settings here at all?
 
 To <dfn algorithm for="PrivateAggregation">consume budget if permitted</dfn> given a {{long}} |value|, an [=origin=] <var ignore=''>origin</var>, a [=context type=] |api| and a [=moment=] |currentTime|, perform [=implementation-defined=] steps. They return a [=boolean=], which indicates whether there is sufficient 'contribution budget' left to send the requested contribution |value|. This budget should be bound to usage over time, e.g. the contribution sum over the last 24 hours. The algorithm should assume that the contribution will be sent if and only if true is returned, i.e. it should consume the budget in that case.
 
-To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=origin=] |reportingOrigin|, a [=context type=] |api|, a [=list=] of {{PAHistogramContribution}} |contributions| and a [=moment=] |currentTime|, perform the following steps. They return an [=aggregatable report=].
+To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=origin=] |reportingOrigin|, a [=context type=] |api|, a [=list=] of {{PAHistogramContribution}}s |contributions| and a [=moment=] |currentTime|, perform the following steps. They return an [=aggregatable report=].
 1. [=Assert=]: |reportingOrigin| is a [=potentially trustworthy origin=].
 1. Let |reportTime| be the result of running [=PrivateAggregation/obtain a report delivery time=] given |currentTime|.
 1. Let |report| be a new [=aggregatable report=] with the items:
@@ -320,6 +320,7 @@ To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=
     :: The result of [=generating a random UUID=].
     : [=aggregatable report/queued=]
     :: false
+1. Return |report|.
 
 Issue: Go through and be consistent about namespacing of algorithms.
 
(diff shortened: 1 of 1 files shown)


alexmturner:
@linnan-github, could you PTAL here too? This mostly just copies what the ARA spec already has, with some modifications/adaptions. Thanks!

spec.bs
```diff
@@ -247,35 +260,55 @@ A user agent holds an <dfn>aggregatable report cache</dfn>, which is a [=list=]
...
-1. If |scope|'s [=worklet global scope type=] is {{FledgeWorkletGlobalScope}}, then return "<code>[=context type/fledge=]</code>".
-1. [=Assert=]: |scope|'s [=worklet global scope type=] is {{SharedStorageWorkletGlobalScope}}.
-1. Return "<code>[=context type/shared-storage=]</code>". -->
+Issue: This would ideally be replaced by a more descriptive algorithm in Infra. See infra/201
 
-To <dfn export algorithm for="PrivateAggregation">obtain a reporting endpoint</dfn> given an [=origin=] |reportingOrigin| and [=context type=] |api|, perform the following steps. They return a [=URL=].
+Scheduling reports {#scheduling-reports}
+----------------------------------------
+
+To perform the <dfn algorithm for="PrivateAggregation">report scheduling steps</dfn> with an [=environment settings object=] |context|, a [=context type=] |api| and a [=list=] of {{PAHistogramContribution}}s |contributions|:
```
linnan-github, line 283: It seems a bit confusing that this algorithm is called "report scheduling steps" as it creates the report as well. Wdyt if we call it "report creation and scheduling steps?

spec.bs
```diff
@@ -247,35 +260,55 @@ A user agent holds an <dfn>aggregatable report cache</dfn>, which is a [=list=]
...
+1. Let |reportingOrigin| be |context|'s [=environment settings object/origin=].
 1. [=Assert=]: |reportingOrigin| is a [=potentially trustworthy origin=].
-1. Let |path| be the [=string/concatenation=] of «"<code>.well-known/private-aggregation/report-</code>", |api|».
-1. Let |base| be the result on running the [=URL parser=] on the [=serialization of an origin|serialization=] of |reportingOrigin|.
-1. [=Assert=]: |base| is not failure.
-1. Let |result| be the result of running the [=URL parser=] on |path| with |base|.
-1. [=Assert=]: |result| is not failure.
-1. Return |result|.
+1. If the user agent's [=aggregatable report cache=] [=set/size=] is larger than [=maximum stored reports=], return.
```
linnan-github, line 286: Note that ARA has removed this check in the spec. Shall we align given that this is not implemented as well for PAA?

spec.bs
```diff
@@ -247,35 +260,55 @@ A user agent holds an <dfn>aggregatable report cache</dfn>, which is a [=list=]
...
+
+Issue: Do we need to ensure the reports aren't queued after being sent?
 
-Issue: Remove export when this definition is used.
+Issue: Do we need to address user settings here at all?
 
-To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=origin=] |reportingOrigin|, a [=context type=] |api| and a [=list=] of {{PAHistogramContribution}} |contributions|, perform the following steps. They return an [=aggregatable report=].
+Issue: Does the [=maximum stored reports=] part match the implementation?
+
+To <dfn algorithm for="PrivateAggregation">consume budget if permitted</dfn> given a {{long}} |value|, an [=origin=] <var ignore=''>origin</var>, a [=context type=] |api| and a [=moment=] |currentTime|, perform [=implementation-defined=] steps. They return a [=boolean=], which indicates whether there is sufficient 'contribution budget' left to send the requested contribution |value|. This budget should be bound to usage over time, e.g. the contribution sum over the last 24 hours. The algorithm should assume that the contribution will be sent if and only if true is returned.
```
linnan-github, line 307: This algorithm doesn't really consume the budget. Maybe call it "check if there is sufficient budget"?

spec.bs
```diff
@@ -289,46 +322,179 @@ To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=
...
+        Note: It might be more practical to perform these steps when the [=user agent=] next starts up.
+
+To <dfn>attempt to deliver a report</dfn> given an [=aggregatable report=] |report|:
+1. Let |url| be the result of [=obtaining a reporting endpoint=] given |report|'s [=aggregatable report/reporting origin=] and |report|'s [=aggregatable report/api=].
+1. Let |data| be the result of [=serializing an aggregatable report=] given |report|.
+1. If |data| is an error, [=list/remove=] |report| from the [=aggregatable report cache=]. (TODO: do we need to queue this task?)
+1. Let |request| be the result of [=creating a report request=] given |url| and |data|.
+1. [=Queue a task=] to [=fetch=] |request| with [=fetch/processResponse=] being the following steps that take a [=response=] |response|:
+    1. Let |shouldRetry| be false.
+    1. If |response|'s [=response/status=] is not 200, optionally set |shouldRetry| to true.
```
linnan-github, line 369: Do we need to call this out explicitly, or just say in case of error?

spec.bs
alexmturner: Sounds good, changed.

spec.bs
alexmturner: Yep, removed.

spec.bs
alexmturner: The idea is that this algorithm does actually consume the budget, but leaves what that means as implementation-defined. Tried to reword slightly to clarify, but lmk what you think

spec.bs
alexmturner: I think I'd lean towards either leaving it as is or making |shouldRetry| completely implementation-defined. (Just not sure "error" by itself is clearly defined here.)

spec.bs
linnan-github: I see. Yeah it's clearer with the rewording, thanks.

spec.bs
```diff
@@ -248,34 +261,50 @@ A user agent holds an <dfn>aggregatable report cache</dfn>, which is a [=list=]
...
-1. Return |result|.
+1. Let |truncatedContributions| be a new [=list/is empty|empty=] [=list=].
+1. If |contributions| has a [=list/size=] greater than [=maximum report contributions=]:
+    1. [=set/For each=] |n| of [=the exclusive range|the range=] 0 to [=maximum report contributions=], exclusive:
+        1. [=set/Append=] |contributions|[|n|] to |truncatedContributions|.
+1. Otherwise, set |truncatedContributions| to |contributions|.
+1. Let |contributionSum| be 0.
+1. [=set/iterate|For each=] |contribution| of |truncatedContributions|:
+    1. [=Assert=]: |contribution|["|value|"] is non-negative.
+    1. Add |value| to |contributionSum|.
```
linnan-github, line 293:
```suggestion
    1. Add |contribution|["|value|"] to |contributionSum|.
```

spec.bs
```diff
@@ -248,34 +261,50 @@ A user agent holds an <dfn>aggregatable report cache</dfn>, which is a [=list=]
...
+
+To <dfn algorithm for="PrivateAggregation">consume budget if permitted</dfn> given a {{long}} |value|, an [=origin=] <var ignore=''>origin</var>, a [=context type=] |api| and a [=moment=] |currentTime|, perform [=implementation-defined=] steps. They return a [=boolean=], which indicates whether there is sufficient 'contribution budget' left to send the requested contribution |value|. This budget should be bound to usage over time, e.g. the contribution sum over the last 24 hours. The algorithm should assume that the contribution will be sent if and only if true is returned, i.e. it should consume the budget in that case.
 
-To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=origin=] |reportingOrigin|, a [=context type=] |api| and a [=list=] of {{PAHistogramContribution}} |contributions|, perform the following steps. They return an [=aggregatable report=].
+To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=origin=] |reportingOrigin|, a [=context type=] |api|, a [=list=] of {{PAHistogramContribution}} |contributions| and a [=moment=] |currentTime|, perform the following steps. They return an [=aggregatable report=].
```
linnan-github, line 305:
```suggestion
To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=origin=] |reportingOrigin|, a [=context type=] |api|, a [=list=] of {{PAHistogramContribution}}s |contributions| and a [=moment=] |currentTime|, perform the following steps. They return an [=aggregatable report=].
```

spec.bs
```diff
@@ -289,46 +318,179 @@ To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=
     :: |api|
     : [=aggregatable report/report ID=]
     :: The result of [=generating a random UUID=].
+    : [=aggregatable report/queued=]
+    :: false
```
linnan-github, line 323:
```suggestion
1. Return |report|.

```

spec.bs
```diff
@@ -289,46 +318,179 @@ To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=
...
+To <dfn>attempt to queue reports for sending</dfn> given a [=list=] of [=aggregatable reports=] |reports|:
+1. [=list/iterate|For each=] |report| of |reports|, run these steps [=in parallel=]:
+    1. Run these steps, but [=abort when=] the [=user agent=] shuts down:
+        1. If |report|'s [=aggregatable report/queued=] value is true, return.
+        1. Set |report|'s [=aggregatable report/queued=] value to true.
+        1. If |report|'s [=aggregatable report/report time=] is less than the [=wall clock=]'s [=current unsafe time=], set |report|'s [=aggregatable report/report time=] to [=wall clock=]'s [=current unsafe time=] plus an [=implementation-defined=] random non-negative [=duration=].
 
-Issue: Don't we need to take into account the time "now" as well?
+            Note: On startup, it is possible the user agent will need to send many reports whose report times passed while the browser was closed. Adding random delay prevents temporal joining of reports.
+        1. Wait until the [=wall clock=]'s [=current unsafe time=] is equal to |report|'s [=aggregatable report/report time=].
```
linnan-github, line 349: Is it ok to compare equality for a coarsen time and an unsafe time?

spec.bs
linnan-github: Maybe just make |shouldRetry| implementation-defined to avoid ambiguity?

spec.bs
```diff
@@ -289,46 +318,179 @@ To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=
...
+    : "`api`"
+    :: |report|'s [=aggregatable report/api=]
+    : "`report_id`"
+    :: |report|'s [=aggregatable report/report ID=]
+    : "`reporting_origin`"
+    :: The [=serialization of an origin|serialization=] of |report|'s [=aggregatable report/reporting origin=]
+    : "`scheduled_report_time`"
+    :: The number of seconds in |scheduledReportTime|, rounded down to the nearest number of whole seconds and [=serialize an integer|serialized=]
+    : "`version`"
+    :: "`0.1`"
```
linnan-github, line 490: We should not hardcode this right?

alexmturner: I was thinking we might want to -- if we're updating the version, that means the spec would have changed, right?

linnan-github: Hmmm, but does it have to be the same for different implementations? I would slightly lean towards making it less explicit. Wdyt?

alexmturner: Yeah, this is an interesting one. We say that obtaining the public key is implementation-defined, which would allow for different aggregation services to be used. But aligning on version should be possible (if one aggregation service needs to make a breaking change, they're presumably already no longer spec compliant). That being said, probably not too critical either way.

@csharrison any thoughts on this?

spec.bs
alexmturner: Done!

spec.bs

spec.bs

spec.bs
alexmturner: Went with that, but added a "should" sentence -- I think it not being a requirement makes me more okay with the definition of an error being potentially uncertain, but lmk what you think :)

spec.bs
alexmturner: My reading is it might be, but we may as well change to "current wall time" anyway.

Two follow up questions: should we change all of these from |context|'s [=current wall time=] to just the general [=current wall time=]? I'm also curious if the clock we're using (i.e. base::Time::Now()) actually does this coarsening. Do you know?

side note: I've changed this to "equal to or after", but lmk if you preferred it before

spec.bs
csharrison: I think speccing the version makes sense to me, if we specify the version as having a 1:1 correspondence with the data format from the client.

spec.bs
linnan-github: @apasel422 made the change for the wall time for ARA and should know more about this. @apasel422 Do you know whether it's preferred to use context's current wall time or the general current wall time? ARA uses context's current wall time throughout. Thanks.

apasel422: No, I don't know.

spec.bs
```diff
@@ -289,46 +318,180 @@ To <dfn for="PrivateAggregation">obtain an aggregatable report</dfn> given an [=
...
+    :   [=request/method=]
+    ::  "`POST`"
+    :   [=request/URL=]
+    ::  |url|
+    :   [=request/header list=]
+    ::  «("`Content-Type`", "`application/json`")»
+    :   [=request/unsafe-request flag=]
+    ::  set
+    :   [=request/body=]
+    ::  |body|
```
linnan-github, line 390: ARA uses "A [body](https://fetch.spec.whatwg.org/#concept-body) whose [source](https://fetch.spec.whatwg.org/#concept-body-source) is body.", shall we aligh?

spec.bs
linnan-github: The context's current wall time may have higher precision in [cross-origin-isolated contexts](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-cross-origin-isolated-capability). Does this apply to shared storage or fledge worklet? If so, maybe better to use the context's current wall time.

alexmturner: Just given that we use base::Time::Now() throughout, my instinct would be to match that 'global' state in the spec (unless we also want to change the implementation)

spec.bs
alexmturner: The fetch spec defines the body as null, a byte sequence or a body, so I'd lean slightly towards just passing the byte sequence in and letting the fetch spec do the rest. Wdyt?

spec.bs
linnan-github: Yeah I'm fine with that, just keeping it consistent in the spec.

alexmturner: Done. I'll try and follow up with the spec mentor about this too

spec.bs
alexmturner: Thanks! Planning to leave as is then :)

spec.bs
```diff
@@ -25,9 +25,20 @@ urlPrefix: https://wicg.github.io/shared-storage/; type: interface
     text: SharedStorageWorklet
     text: SharedStorageWorkletGlobalScope
 spec: hr-time; type: dfn; urlPrefix: https://w3c.github.io/hr-time/
+    text: current unsafe time; url: #wall-clock-unsafe-current-time
```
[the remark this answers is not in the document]
alexmturner: Nope, removed.

spec.bs
```diff
@@ -248,34 +260,49 @@ A user agent holds an <dfn>aggregatable report cache</dfn>, which is a [=list=]
...
+1. Let |truncatedContributions| be a new [=list/is empty|empty=] [=list=].
+1. If |contributions| has a [=list/size=] greater than [=maximum report contributions=]:
+    1. [=set/For each=] |n| of [=the exclusive range|the range=] 0 to [=maximum report contributions=], exclusive:
+        1. [=set/Append=] |contributions|[|n|] to |truncatedContributions|.
+1. Otherwise, set |truncatedContributions| to |contributions|.
+1. Let |contributionSum| be 0.
+1. [=set/iterate|For each=] |contribution| of |truncatedContributions|:
+    1. [=Assert=]: |contribution|["|value|"] is non-negative.
+    1. Add |contribution|["|value|"] to |contributionSum|.
+1. Let |sufficientBudget| be the result of [=PrivateAggregation/consuming budget if permitted=] given |contributionSum|, |reportingOrigin|, |api| and the [=current wall time=].
```
linnan-github, line 292: nit: Shall we put current wall time in a variable and used later?

linnan-github:
LGTM!

spec.bs
alexmturner: Makes sense! Done

alexmturner:
Thanks for all the reviews!

PR #38, opened 2023-05-11
Limit number of contributions to 20 per report

### LINKED ISSUE #32: Consider reducing maximum number of contributions per report
We plan to [pad the encrypted payload](https://github.com/patcg-individual-drafts/private-aggregation-api#padding), which will increase the size of the payload. To mitigate this size increase, we're considering reducing the number of contributions allowed in each report (e.g. from 50 to 20).

Note also that currently, any additional contributions would be dropped, but we could also consider allowing batching of contributions into multiple reports if this maximum value is limiting. Please let us know if you have any thoughts!

alexmturner:
Resolves #32.

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,8 @@ we should truncate them to the limit.
 If necessary, we could instead split the contributions back into multiple
 reports, each respecting the limit.
 
+Strawman limit: 20 contributions per report.
+
 #### Padding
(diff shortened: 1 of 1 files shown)


alexmturner:

linnan-github:
LGTM!

PR #50, opened 2023-05-24
Fix debug reporting path

alexmturner:
The `debug/` infix is placed in the wrong spot currently.


alexmturner:
@linnan-github, PTAL when you have time :) Thanks!

linnan-github (approved):

PR #51, opened 2023-05-30
Change API layering in spec

alexmturner:
Creates new exported algorithms and one algorithm to be overriden. This cl doesn't actually move any content to other specs, but is a step to making that simpler. This change also addresses batching, by defining the batching scopes more completely.

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/51.html" title="Last updated on Jun 9, 2023, 7:47 PM UTC (ada0439)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/51/b69953e...ada0439.html" title="Last updated on Jun 9, 2023, 7:47 PM UTC (ada0439)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -794,9 +794,9 @@ A signal base value is one of the following:
 : "<dfn export><code>highest-scoring-other-bid</code></dfn>"
 :: The bid value of the highest scoring bid that did not win.
 : "<dfn export><code>script-run-time</code></dfn>"
-:: The running time of the script in ms(?).
+:: The running time of the script in milliseconds.
 : "<dfn export><code>signals-fetch-time</code></dfn>"
-:: The time it took for the signals fetch to complete in ms(?)
+:: The time it took for the signals fetch to complete in milliseconds.
 : "<dfn export><code>bid-reject-reason</code></dfn>"
 :: The reason a bid was rejected.
 
--- a/spec.bs
+++ b/spec.bs
@@ -157,13 +157,30 @@ global scope.
 
 Additionally, each global scope should set the [=PrivateAggregation/get batching
 scope steps=] for the {{PrivateAggregation}} object it exposes to a non-null
-value. The global scope may wait to set the field to indicate that the API is
-not yet available, for example, during the initial execution of a script after
-loading. For any batching scope returned by the [=PrivateAggregation/get
-batching scope steps=], the [=mark a batching scope complete=] steps should
-later be performed given that same batching scope, the global scope's [=relevant
-settings object=]'s [=environment settings object/origin=] and some [=context
-type=].
+value. The global scope should wait to set the field until the API is intended
+to be available.
+
+<div class="example" id="shared-storage-only-within-operations">
+    Shared Storage only allows Private Aggregation when an operation is being
+    invoked, not in the top-level context:
+    <pre highlight="js">
+    class ExampleOperation {
+      async run(data) {
+        privateAggregation.contributeToHistogram(...)  // This is allowed.
+      }
+    }
+    register('example-operation', ExampleOperation);
+
+    privateAggregation.contributeToHistogram(...)  // This would cause an error.
+    </pre>
+    So, Shared Storage sets [=PrivateAggregation/get batching scope steps=]
+    after the initial execution of the module script is complete.
+</div>
+
+For any batching scope returned by the [=PrivateAggregation/get batching scope
+steps=], the [=mark a batching scope complete=] steps should later be performed
+given that same batching scope, the global scope's [=relevant settings
+object=]'s [=environment settings object/origin=] and some [=context type=].
 
 Note: This last requirement means that global scopes with different origins
     cannot share the same batching scope.
--- a/spec.bs
+++ b/spec.bs
@@ -133,6 +133,8 @@ are:
     [=PrivateAggregation/get batching scope steps=].
 1. If |batchingScopeSteps| is null, throw a new {{DOMException}} with name
     "`NotAllowedError`".
+    Note: This indicates the API is not yet available.
+
 1. Let |entry| be a new [=contribution cache entry=] with the items:
     : [=contribution cache entry/contribution=]
     :: |contribution|
@@ -144,7 +146,6 @@ are:
 Issue: Check that value can actually be zero in the spec pipeline.
 
 Issue(44): Consider accepting an array of contributions.
-
 
 Exposing to global scopes {#exposing}
 =====================================
@@ -604,8 +605,6 @@ When an operation invocation completes, [=mark a batching scope complete=] given
 the associated [=batching scope=], the global scope's [=relevant settings
 object=]'s [=environment settings object/origin=] and
 "<code>shared-storage</code>".
-
-Issue: How do we handle overrides without causing bikeshed errors?
 
 Protected Audience API monkeypatches {#protected-audience-api-monkeypatches}
 ============================================================================
--- a/spec.bs
+++ b/spec.bs
@@ -609,8 +609,8 @@ To <dfn>obtain a report's shared info</dfn> given an [=aggregatable report=]
 1. Return the result of [=serializing an infra value to a json string=] given
     |sharedInfo|.
 
-Shared Storage API monkeypatches {#shared-storage-api-monkeypatches}
-====================================================================
+Shared Storage API monkey patches {#shared-storage-api-monkey-patches}
+======================================================================
 
 Issue(43): This should be moved to the Shared Storage spec.
 
@@ -624,12 +624,12 @@ The {{WindowSharedStorage}}'s {{WindowSharedStorage/run()}} method steps are
 modified in two ways. First, add the following step in the nested scope just
 after "Let |operation| be |operationMap|[|name|]." (renumbering later steps as
 appropriate):
-<div algorithm="shared-storage-run-monkeypatch-1">
+<div algorithm="shared-storage-run-monkey-patch-1">
 2. Let <var ignore>batchingScope</var> be a new [=batching scope=].
 
 </div>
 Second, at the end of the same nested scope, add the following step:
-<div algorithm="shared-storage-run-monkeypatch-2">
+<div algorithm="shared-storage-run-monkey-patch-2">
 5. When the above [=call=] returns, [=process contributions for a batching
     scope=] given <var ignore>batchingScope</var>,
     <var ignore>outsideSettings</var>' [=environment settings object/origin=]
@@ -641,12 +641,12 @@ The {{WindowSharedStorage}}'s {{WindowSharedStorage/selectURL()}} method steps
 are modified in two ways. First, add the following step in the nested scope just
 after "Let |operation| be |operationMap|[|name|]." (renumbering later steps as
 appropriate):
-<div algorithm="shared-storage-selecturl-monkeypatch-1">
+<div algorithm="shared-storage-selecturl-monkey-patch-1">
 2. Let <var ignore>batchingScope</var> be a new [=batching scope=].
 
 </div>
 Second, at the end of the same nested scope, add the following step:
-<div algorithm="shared-storage-selecturl-monkeypatch-2">
+<div algorithm="shared-storage-selecturl-monkey-patch-2">
 8. [=Process contributions for a batching scope=] given
     <var ignore>batchingScope</var>, <var ignore>outsideSettings</var>'
     [=environment settings object/origin=] and "<code>shared-storage</code>".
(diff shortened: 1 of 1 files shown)


alexmturner:
@yoavweiss could you take a quick look? I wasn't sure if this was the best direction towards fixing the layering, but I'm trying to allow each implementing API to define its own 'batching scope'. Specifically, I wasn't really sure whether functions being 'overridden' was a well-supported spec concept. Thanks!

alexmturner:
I've reworked this to separate out the parts that will (eventually) be moved to other specs into separate sections at the end as well as moved from 'overrides' to having a field that gets set by implementing specs

cc @linnan-github

spec.bs
```diff
@@ -130,110 +129,65 @@ are:
     is not in the range [0, 2<sup>128</sup>−1].
 1. Throw {{RangeError}} if |contribution|["{{PAHistogramContribution/value}}"]
     is negative.
-1. [=list/Append=] |contribution| to the [=contributions cache=].
+1. Let |batchingScopeSteps| be the {{PrivateAggregation}} object's
+    [=PrivateAggregation/get batching scope steps=].
+1. If |batchingScopeSteps| is null, throw a new {{DOMException}} with name
```
yoavweiss, line 134: Can that happen? If not, it's better to assert it doesn't, rather than throwing

spec.bs
```diff
@@ -130,110 +129,65 @@ are:
...
 
 Structures {#structures}
 ========================
 
 General {#general-structures}
 -----------------------------
 
-<h4 dfn-type=dfn>Aggregatable report</h3>
+<h4 dfn-type=dfn>Batching scope</h4>
+A batching scope is a unique internal value that identifies which
```
yoavweiss, line 175: s/unique internal value/implementation-defined value/

https://infra.spec.whatwg.org/#implementation-defined

spec.bs
```diff
@@ -632,9 +575,139 @@ To <dfn>obtain a report's shared info</dfn> given an [=aggregatable report=]
...
+
+Note: Multiple operation invocations can be in-progress at the same time, each
+with a different batching scope. However, only one can be currently executing.
+
+When an operation invocation completes, [=mark a batching scope complete=] given
+the associated [=batching scope=], the global scope's [=relevant settings
+object=]'s [=environment settings object/origin=] and
+"<code>shared-storage</code>".
+
+Issue: How do we handle overrides without causing bikeshed errors?
```
yoavweiss, line 608: What overrides?

yoavweiss (commented):
Looks good overall. A few comments

spec.bs
alexmturner: Added a note here -- below we say "The global scope may wait to set the field to indicate that the API is not yet available."

spec.bs
alexmturner: For this, I was trying to emulate the "opaque origin" kind of concept. I saw "unique internal value" in the HTML spec (https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#unique-values), but I don't think it's exported, so I might need to move that definition in here.

spec.bs
alexmturner: This is stale -- removed

alexmturner:

spec.bs
yoavweiss: @domenic - thoughts? Should we try to export "unique internal value"?

spec.bs
alexmturner: This is really just to prevent the API being available during the initial execution of a script (e.g. while `addModule()` runs). We only want the API to be available after that point. Definitely open to suggestions for better structuring this, but for now I've updated the description slightly to clarify.

alexmturner: Added an example to better illustrate this, lmk if you think it's clear enough

spec.bs
```diff
@@ -130,110 +129,85 @@ are:
...
-
-<xmp class="idl">
-partial interface InterestGroupScriptRunnerGlobalScope {
-  readonly attribute PrivateAggregation privateAggregation;
-};
+1. Let |batchingScopeSteps| be the {{PrivateAggregation}} object's
+    [=PrivateAggregation/get batching scope steps=].
+1. If |batchingScopeSteps| is null, throw a new {{DOMException}} with name
+    "`NotAllowedError`".
+    Note: This indicates the API is not yet available, for example, because the
```
yoavweiss, line 136: Style nit: try adding an extra empty line between the note and the algorithm step. As is, it's kinda mashed together..

spec.bs
yoavweiss: It's fine as far as the spec go, but worthwhile thinking about the developer ergonomics here:
* How would developers know when it's fine to run this API call and when it isn't? Is that part of our developer documentation?
* If I'm a library provider enabling folks to send reports, how should I write my code so that it never runs too early? (regardless of when callers may call it)

spec.bs
domenic: Huh I didn't realize that we made that into a whole concept with a definition. Probably we should move that to Infra and make it exported.

In the meantime, you could do `<a spec=HTML>unique internal value</a>` to link to the concept despite it not being exported.

alexmturner: Thanks, moved to the <a spec=HTML> tag as suggested. Also added an issue and filed https://github.com/whatwg/infra/issues/583 to track this.

spec.bs
```diff
@@ -632,9 +601,137 @@ To <dfn>obtain a report's shared info</dfn> given an [=aggregatable report=]
...
+{{SharedStorageWorkletGlobalScope}}, its [=PrivateAggregation/get batching scope
+steps=] should be left default until the `addModule()` execution is complete. At
+that point, it should be set to the following algorithm:
+1. Return the [=batching scope=] associated with the currently executing
+    operation invocation in |scope|.
+
+Note: Multiple operation invocations can be in-progress at the same time, each
+with a different batching scope. However, only one can be currently executing.
+
+When an operation invocation completes, [=mark a batching scope complete=] given
```
yoavweiss, line 629: Same here - is there an algorithm that runs the operation invocation, that can call this logic?

spec.bs
```diff
@@ -632,9 +601,137 @@ To <dfn>obtain a report's shared info</dfn> given an [=aggregatable report=]
...
+partial interface PrivateAggregation {
+  undefined contributeToHistogramOnEvent(
+      DOMString event, PAExtendedHistogramContribution contribution);
+};
+</xmp>
+
+Issue: Do we want to align naming with implementation?
+
+At the start of each Protected Audience auction (i.e. a call to
+`runAdAuction()`), create a new [=batching scope=] for each unique origin
```
yoavweiss, line 666: Can you state that in terms of an algorithm step? (e.g. before step 1 of https://wicg.github.io/turtledove/#dom-navigator-runadauction)

spec.bs
```diff
@@ -632,9 +601,137 @@ To <dfn>obtain a report's shared info</dfn> given an [=aggregatable report=]
...
+participating in the auction (including both bidders and the seller).
+
+When a {{PrivateAggregation}} object is exposed on a
```
yoavweiss, line 669: Same comment as for SharedStorage

alexmturner:
@yoavweiss Thanks for the suggestions, had a go reworking to better integrate with the specs. Ran into a few issues with the other specs, so opened issues and noted those in this spec too where relevant. Some of the integrations I've put here are a little more handwavy than I'd like, but that should be fixable once those issues are resolved.

spec.bs
```diff
@@ -130,110 +131,91 @@ are:
...
-Exposing to the Shared Storage API {#shared-storage}
-====================================================
-
-<xmp class="idl">
-partial interface SharedStorageWorkletGlobalScope {
-  readonly attribute PrivateAggregation privateAggregation;
-};
-</xmp>
+    Note: This indicates the API is not yet available, for example, because the
+        the initial execution of the script after loading is not complete.
```
linnan-github, line 140:
```suggestion
        initial execution of the script after loading is not complete.
```

spec.bs
```diff
@@ -130,110 +131,91 @@ are:
...
+    :: |contribution|
+    : [=contribution cache entry/batching scope=]
+    :: The result of running |batchingScopeSteps|
 
-To <dfn>process the Shared Storage contributions cache</dfn> given a [=list=]
-|contributionsCache| and a {{SharedStorageWorkletGlobalScope}} |scope|, perform
-the [=PrivateAggregation/report creation and scheduling steps=] with |scope|'s
-[=relevant settings object=]'s [=environment settings object/origin=],
-"<code>[=context type/shared-storage=]</code>" and |contributionsCache|.
+1. [=Append an entry to the contribution cache=] given |entry|.
```
linnan-github, line 151: Maybe we can do something like
1. [=Append|Append an entry to the contribution cache=] |entry| to the [=contribution cache=].

spec.bs
```diff
@@ -130,110 +131,91 @@ are:
...
-
-<div algorithm>
-The <dfn method for="PrivateAggregation">contributeToHistogramOnEvent(DOMString
-event, PAExtendedHistogramContribution contribution)</dfn> method steps are:
+Issue(44): Consider accepting an array of contributions.
+
+Exposing to global scopes {#exposing}
+=====================================
+
+To expose this API to a global scope, a readonly attribute of type
```
linnan-github, line 160: Probably can make it a bit more precise like `a readonly attribute privateAggregation of type {{Private Aggregation}}`

spec.bs
```diff
@@ -130,110 +131,91 @@ are:
...
-1. Perform the [=PrivateAggregation/report creation and scheduling steps=] with
-    |scope|'s [=relevant settings object=]'s [=environment settings object/
-    origin=], "<code>[=context type/fledge=]</code>" and
-    |filledInContributions|.
+For any batching scope returned by the [=PrivateAggregation/get batching scope
+steps=], the [=mark a batching scope complete=] steps should later be performed
+given that same batching scope, the global scope's [=relevant settings
+object=]'s [=environment settings object/origin=] and some [=context type=].
 
+Note: This last requirement means that global scopes with different origins
```
linnan-github, line 191: Probably can mention or link to the Same-Origin policy?

spec.bs
```diff
@@ -325,7 +274,29 @@ To <dfn>serialize an integer</dfn>, represent it as a [=string=] of the shortest
...
 
 Issue: This would ideally be replaced by a more descriptive algorithm in Infra.
-    See infra/201
+    See <a href="https://github.com/whatwg/infra/issues/201">infra/201</a>.
+
+To <dfn algorithm export>append an entry to the contribution cache</dfn> given a
+[=contribution cache entry=] |entry|:
+1. [=list/Append=] |entry| to the [=contribution cache=].
+
+To <dfn algorithm export>mark a batching scope complete</dfn> given a [=batching
```
linnan-github, line 283: I wonder whether we should make it more clear in the algorithm name that a report may be created and scheduled.

spec.bs

spec.bs

spec.bs

spec.bs
```diff
@@ -325,7 +274,29 @@ To <dfn>serialize an integer</dfn>, represent it as a [=string=] of the shortest
...
+1. Let |batchEntries| be a new [=list/is empty|empty=] [=list=].
+1. [=list/iterate|For each=] |entry| of the [=contribution cache=]:
+    1. If |entry|'s [=contribution cache entry/batching scope=] is
+        |batchingScope|, [=list/append=] |entry| to |batchEntries|.
+1. Let |batchContributions| be a new [=list/is empty|empty=] [=list=].
+1. [=list/iterate|For each=] |entry| of |batchEntries|:
+    1. [=list/Remove=] |entry| from the [=contribution cache=].
+    1. [=list/Append=] |entry|'s [=contribution cache entry/contribution=] to
+        |batchContributions|.
+1. If |batchEntries| is [=list/empty=], return.
```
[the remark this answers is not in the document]

spec.bs
alexmturner: Went with "Process contributions for a batching scope"

linnan-github (approved):

spec.bs
```diff
@@ -632,9 +609,207 @@ To <dfn>obtain a report's shared info</dfn> given an [=aggregatable report=]
...
+
+<h4 dfn-type=dfn>Signal base value</h3>
+A signal base value is one of the following:
+<dl dfn-for="signal base value">
+: "<dfn export><code>winning-bid</code></dfn>"
+:: The bid value of the winning bid.
+: "<dfn export><code>highest-scoring-other-bid</code></dfn>"
+:: The bid value of the highest scoring bid that did not win.
+: "<dfn export><code>script-run-time</code></dfn>"
+:: The running time of the script in ms(?).
```
linnan-github, line 797: Could we add an Issue to confirm this?

alexmturner: Just confirmed -- removed the uncertainty here

alexmturner:
Thanks for all the improvements!

PR #55, opened 2023-06-08
Use debug_key in JSON report

alexmturner:
We shouldn't have updated "debug_key" to "debugKey" in the JSON report itself -- only in the javascript.


alexmturner:

linnan-github (approved):

Issue #43, opened 2023-05-16
Reliant specs and their layering


yoavweiss:
Reading through the spec, this seems like infrastructure that a couple of other specs are currently relying on, but one that could be potentially useful for future cases. (e.g. one can imagine a "cross-origin data worklet" that has access to cross-origin data and uses it to generate performance-related histograms in a secure way that doesn't leak any user-specific data)

Given that, I think it'd make sense to make this spec more agnostic to specs that rely on it. In order to do that, I think it'd make sense to move all the SharedStrorage/FLEDGE specific bits to their respective specifications, and only define the infrastructure (that can serve both cases) here.

alexmturner:
Note this has been partially addressed by #51, but leaving the issue open until the monkey patches are moved to the Shared Storage/Protected Audience specs

PR #63, opened 2023-06-14
Spec: Fill in Protected Audience extensions

alexmturner:
Fills in many missing pieces from the Protected Audience monkey patches

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/63.html" title="Last updated on Jun 20, 2023, 7:39 PM UTC (e53f1bd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/63/f2a110f...e53f1bd.html" title="Last updated on Jun 20, 2023, 7:39 PM UTC (e53f1bd)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1117,11 +1117,26 @@ Issue: Once <a href="https://github.com/wicg/turtledove/issues/615">
 Protected Audience API-specific structures {#protected-audience-api-specific-structures}
 ----------------------------------------------------------------------------------------
 
+<h4 id="extending-auction-config">Extending auction config</h4>
+
+Extend the <a spec="turtledove">auction config</a> [=struct=] to add a new
+field:
+: <dfn for="auction config" spec="turtledove">per-bid or seller on event
+    contribution cache</dfn>
+:: A [=map=] from [=interest group=] or null to a [=on event contribution cache=].
+: <dfn for="auction config" spec="turtledove">batching scope map</dfn>
+:: A [=map=] from [=origin=] to [=batching scope=], initially
+    [=map/is empty|empty=].
+
+    Note: Does not include [=batching scopes=] for contributions conditional on
+        non-reserved events.
+
+<br/>
 Add the following definitions in a new subsection at the end of
 <a href="https://wicg.github.io/turtledove/#structures">Structures</a>,
 renumbered appropriately.
 
-<h4 dfn-type=dfn>Signal base value</h3>
+<h4 dfn-type=dfn>Signal base value</h4>
 A signal base value is one of the following:
 <dl dfn-for="signal base value">
 : "<dfn><code>winning-bid</code></dfn>"
@@ -1179,18 +1194,6 @@ An on event contribution cache entry is a [=struct=] with the following items:
 <h4 dfn-type=dfn>On event contribution cache</h4>
 An on event contribution cache is a [=map=] from [=string=] to a [=list=] of
 [=on event contribution cache entries=].
-
-Extend the <a spec="turtledove">auction config</a> [=struct=] to add a new
-field:
-: <dfn for="auction config" spec="turtledove">per-bid or seller on event
-    contribution cache</dfn>
-:: A [=map=] from [=interest group=] or null to a [=on event contribution cache=].
-: <dfn for="auction config" spec="turtledove">batching scope map</dfn>
-:: A [=map=] from [=origin=] to [=batching scope=], initially
-    [=map/is empty|empty=].
-
-    Note: Does not include [=batching scopes=] for contributions conditional on
-        non-reserved events.
 
 Protected Audience API-specific algorithms {#protected-audience-api-specific-algorithms}
 ----------------------------------------------------------------------------------------
--- a/spec.bs
+++ b/spec.bs
@@ -996,7 +996,7 @@ event, PAExtendedHistogramContribution contribution)</dfn> method steps are:
 1. If |bucket| is a {{PASignalValue}}:
     1. If |bucket|["{{PASignalValue/baseValue}}"] is not a valid [=signal base
         value=], throw a {{TypeError}}.
-    1. If |bucket|["{{PASignalValue/offset}}"] is a {{long}}, throw a
+    1. If |bucket|["{{PASignalValue/offset}}"] is not a {{bigint}}, throw a
         {{TypeError}}.
 1. Let |value| be |contribution|["{{PAExtendedHistogramContribution/value}}"].
 1. If |value| is a {{PASignalValue}}:
@@ -1035,6 +1035,11 @@ event, PAExtendedHistogramContribution contribution)</dfn> method steps are:
 1. If |onEventContributionCache|[|event|] does not [=map/exist=], set
     |onEventContributionCache|[|event|] to a new [=list=].
 1. [=list/Append=] |entry| to |onEventContributionCache|[|event|].
+
+Issue: Expose the auction config or per-bid or seller on event contribution
+    cache to each global scope. Also expose the interest group to each
+    {{InterestGroupReportingScriptRunnerGlobalScope}} that is associated with a
+    bidder. This will allow for replacing the handwavy prose above.
 
 Issue(44): Consider accepting an array of contributions.
 
@@ -1125,7 +1130,8 @@ Extend the <a spec="turtledove">auction config</a> [=struct=] to add a new
 field:
 : <dfn for="auction config" spec="turtledove">per-bid or seller on event
     contribution cache</dfn>
-:: A [=map=] from [=interest group=] or null to a [=on event contribution cache=].
+:: A [=map=] from [=interest group=] or null to a [=on event contribution
+    cache=].
 : <dfn for="auction config" spec="turtledove">batching scope map</dfn>
 :: A [=map=] from [=origin=] to [=batching scope=], initially
     [=map/is empty|empty=].
--- a/spec.bs
+++ b/spec.bs
@@ -1333,6 +1333,8 @@ To <dfn>determine a signal's numeric value</dfn> given a [=signal base value=]
     : <code>[=worklet function/report-win=]</code>
     :: Return 0.
 
+    </dl>
+
     Issue: Consider disallowing this in the latter two [=worklet functions=].
 1. If |signalBaseValue| is
     "<code>[=signal base value/bid-reject-reason=]</code>":
--- a/spec.bs
+++ b/spec.bs
@@ -402,6 +402,11 @@ To <dfn algorithm export>append an entry to the contribution cache</dfn> given a
 [=contribution cache entry=] |entry|:
 1. [=list/Append=] |entry| to the [=contribution cache=].
 
+To <dfn algorithm export>get a debug details</dfn> given a [=debug scope=]
+|debugScope:
+1. [=Assert=]: [=debug scope map=][|debugScope|] [=map/exists=]
+1. Return [=debug scope map=][|debugScope|].
+
 To <dfn algorithm export>mark a debug scope complete</dfn> given a [=debug
 scope=] |debugScope|:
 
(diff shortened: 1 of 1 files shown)


alexmturner:
@linnan-github, could you PTAL? Unfortunately, it's a bit handwavy as there are some missing pieces in the Protected Audience spec that would ease integration. But I've noted those cases and filed an issue. Thanks!

cc @yoavweiss

spec.bs
```diff
@@ -381,6 +381,11 @@ To <dfn algorithm export>append an entry to the contribution cache</dfn> given a
 [=contribution cache entry=] |entry|:
 1. [=list/Append=] |entry| to the [=contribution cache=].
 
+To <dfn algorithm export>get a debug details</dfn> given a [=debug scope=]
+|debugScope:
```
linnan-github, line 385:
```suggestion
|debugScope|:
```

spec.bs
```diff
@@ -381,6 +381,11 @@ To <dfn algorithm export>append an entry to the contribution cache</dfn> given a
...
+1. [=Assert=]: [=debug scope map=][|debugScope|] [=map/exists=]
```
linnan-github, line 386:
```suggestion
1. [=Assert=]: [=debug scope map=][|debugScope|] [=map/exists=].
```

alexmturner:
@qingxinwu could you also take a look? Thanks!

spec.bs
```diff
@@ -1101,13 +1303,81 @@ integer |maxAllowed|, perform the following steps. They return an integer.
...
+    : <code>[=worklet function/report-win=]</code>
+    :: Return 0.
+
+    </dl>
+
+    Issue: Consider disallowing this in the latter two [=worklet functions=].
+1. If |signalBaseValue| is
+    "<code>[=signal base value/bid-reject-reason=]</code>":
+    1. If the bid did not succeed purely because it didn't meet the required
+        k-anonymity threshold, return 8.
```
qingxinwu, line 1345: maybe useful to export the ["k-anonymity threshold" definition](https://github.com/WICG/turtledove/blob/main/spec.bs#LL1401C43-L1401C43) from Protected Audience spec and add a autolink here using `[=k-anonymity threshold=]`? I can export that if you'd like to link to it.

qingxinwu (commented):
A few quick comments. Still reviewing.
Sorry got distracted by my own PR.

spec.bs
```diff
@@ -964,31 +975,96 @@ partial interface PrivateAggregation {
...
+Issue(44): Consider accepting an array of contributions.
+
+Protected Audience API algorithm modifications {#protected-audience-api-algorithm-modifications}
+------------------------------------------------------------------------------------------------
+
+The <a spec="turtledove">validate and convert auction ad config</a> steps are
+modified to add the following steps just before the last step ("Return
+<var ignore>auctionConfig</var>"), renumbering the later step as appropriate:
+<div algorithm="protected-audience-validate-config-monkey-patch">
+7. Let |batchingScopeMap| be |auctionConfig|'s [=auction config/batching scope
```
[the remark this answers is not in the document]
alexmturner: Yeah, this is a little unfortunate, but I think monkey patching in general is a bit 'fickle'. The prose above is written to be more resilient (putting the text of the step to make it clear even if the numbering changes). Still, we're planning to just move this logic to the other specs without monkey patches soon (see issue #43).

spec.bs
```diff
@@ -964,31 +975,96 @@ partial interface PrivateAggregation {
...
+    [=list=].
+1. [=list/Append=] |entry| to |onEventContributionCache|[|event|].
+
+Issue(44): Consider accepting an array of contributions.
+
+Protected Audience API algorithm modifications {#protected-audience-api-algorithm-modifications}
+------------------------------------------------------------------------------------------------
+
+The <a spec="turtledove">validate and convert auction ad config</a> steps are
+modified to add the following steps just before the last step ("Return
```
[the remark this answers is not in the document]
alexmturner: See comment above, but not planning to monkey patch for too long. This kind of wording is what I've seen other specs do (e.g. https://wicg.github.io/shared-storage/#worklet-monkey-patch, https://wicg.github.io/shared-storage/#worklet-monkey-patch). These descriptions do link to the specific spec algorithm, so the only risk is renumbering of steps. For that, I've included wording from the step itself to clarify. Unfortunately, I don't think spec.bs line numbers are very stable.

spec.bs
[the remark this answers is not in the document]
qingxinwu: ah right, forget that we're going to move them to Protected Audience spec. Then it's totally fine.

spec.bs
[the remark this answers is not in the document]
qingxinwu: same as the other, it's totally fine then.

spec.bs
alexmturner: Added this link :)

spec.bs
```diff
@@ -1101,13 +1303,81 @@ integer |maxAllowed|, perform the following steps. They return an integer.
...
+
+    </dl>
+
+    Issue: Consider disallowing this in the latter two [=worklet functions=].
+1. If |signalBaseValue| is
+    "<code>[=signal base value/bid-reject-reason=]</code>":
+    1. If the bid did not succeed purely because it didn't meet the required
+        k-anonymity threshold, return 8.
+    1. Let |bidRejectReason| be "`not-available`".
+    1. If the seller provided a reject reason, set |bidRejectReason| to that
```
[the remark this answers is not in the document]
alexmturner: Yeah I think we have to wait, but I've added an issue to track this

spec.bs
```diff
@@ -1024,73 +1114,185 @@ Issue: Once <a href="https://github.com/wicg/turtledove/issues/615">
...
+Protected Audience API-specific structures {#protected-audience-api-specific-structures}
+----------------------------------------------------------------------------------------
 
-Issue: Need to document limits on offset, etc.
+<h4 id="extending-auction-config">Extending auction config</h4>
 
-Issue: Fill in the rest. (Need to put the contribution in some sort of queue and
-    process the queue at some point. Need to decide where the queue should live
-    given it has to outlive the auction.)
+Extend the <a spec="turtledove">auction config</a> [=struct=] to add a new
```
qingxinwu, line 1122: Is this implemented. If so, did we save those to auction config?
I think auction config is not something we'd like to modify.
If it's something provided by the browser, how about [browserSignals](https://github.com/WICG/turtledove/blob/main/spec.bs#L1011)?

spec.bs
```diff
@@ -964,31 +975,96 @@ partial interface PrivateAggregation {
...
+    1. If |bucket|["{{PASignalValue/offset}}"] is a {{long}}, throw a
+        {{TypeError}}.
+1. Let |value| be |contribution|["{{PAExtendedHistogramContribution/value}}"].
+1. If |value| is a {{PASignalValue}}:
+    1. If |value|["{{PASignalValue/baseValue}}"] is not a valid [=signal base
+        value=], throw a {{TypeError}}.
+    1. If |value|["{{PASignalValue/offset}}"] is a {{bigint}}, throw a
+        {{TypeError}}.
+1. Let |batchingScope| be null.
+1. If |entry| [=string/starts with=] "`reserved.`", set |batchingScope| to the
```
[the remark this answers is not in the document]
alexmturner: Fixed

spec.bs
```diff
@@ -964,31 +975,96 @@ partial interface PrivateAggregation {
...
+1. If this function execution is associated with a bidder, set |ig| to the
+    [=interest group=] associated with the current function execution.
+1. Let |cacheMap| be |auctionConfig|'s [=auction config/per-bid or seller on
+    event contribution cache=].
+1. If |cacheMap|[|ig|] does not [=map/exist=], [=map/set=] |cacheMap|[|ig|] to
+    a new [=map/is empty|empty=] [=on event contribution cache=].
+1. Let |onEventContributionCache| be |cacheMap|[|ig|].
+1. If |onEventContributionCache|[|event|] does not [=map/exist=], set
+    |onEventContributionCache|[|event|] to a new [=list/is empty|empty=]
+    [=list=].
```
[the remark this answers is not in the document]
alexmturner: Ah thanks -- removed this here and put an issue to fix it elsewhere (will do in a follow up)

spec.bs
alexmturner: I think browserSignals is something that gets passed as a JavaScript argument to bidding script.

I went with extending the auction config as it's currently used to store some per-auction state (e.g. the pending promise count https://wicg.github.io/turtledove/#auction-config-pending-promise-count). I don't think the spec's auction config is exactly 1:1 with the implementation's AuctionConfig (the implementation's doesn't include this state).

I also couldn't really see any other spec concept that stores per-auction state.

qingxinwu: oh I messed up the AuctionConfig IDL and the struct somehow. The struct [=auction config=] seems fine.

spec.bs
```diff
@@ -964,31 +978,95 @@ partial interface PrivateAggregation {
...
+    :: The result of running |scopingDetails|' [=scoping details/get debug scope
+        steps=].
+    : [=on event contribution cache entry/worklet function=]
+    :: The [=worklet function=] that is currently being executed.
+
+1. Let |auctionConfig| be the <a spec="turtledove">auction config</a> associated
+    with the global scope.
+1. Let |ig| be null.
+1. If this function execution is associated with a bidder, set |ig| to the
+    [=interest group=] associated with the current function execution.
```
qingxinwu, line 1029: High level wise, this is fine.
Spec wise, this can be written as something similar to [sendReportTo](
https://wicg.github.io/turtledove/#dom-interestgroupreportingscriptrunnerglobalscope-sendreportto), like `this's relevant global object's's [=InterestGroupBiddingScriptRunnerGlobalScope/interest group=]`

spec.bs
```diff
@@ -1024,73 +1116,185 @@ Issue: Once <a href="https://github.com/wicg/turtledove/issues/615">
...
 
-Issue: Need to document limits on offset, etc.
+<h4 id="extending-auction-config">Extending auction config</h4>
 
-Issue: Fill in the rest. (Need to put the contribution in some sort of queue and
-    process the queue at some point. Need to decide where the queue should live
-    given it has to outlive the auction.)
+Extend the <a spec="turtledove">auction config</a> [=struct=] to add a new
+field:
+: <dfn for="auction config" spec="turtledove">per-bid or seller on event
```
qingxinwu, line 1126: is it better to define these as fields of InterestGroupScriptRunnerGlobalScope, similar to https://github.com/WICG/turtledove/blob/main/spec.bs#L1914?
This seems to be the only things needed from auctionConfig, so we didn't add auctionConfig to GlobalScope, and you cannot retrive auction config from GlobalScope, unless this spec or protected audience spec adds it to GlobalScope.

qingxinwu (commented):
LGTM high level wise.

Left some comments that might need some change spec detail wise.

spec.bs
```diff
@@ -964,31 +978,95 @@ partial interface PrivateAggregation {
 
 Issue: Do we want to align naming with implementation?
```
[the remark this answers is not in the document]
alexmturner: Thanks! Will follow up

spec.bs
```diff
@@ -964,31 +978,95 @@ partial interface PrivateAggregation {
...
+    "`reserved.loss`", "`reserved.win`" » does not [=list/contain=] |event|,
+    return.
+
+    Note: No error is thrown to allow forward compatibility if additional
+        reserved event types are added later.
+1. Let |bucket| be |contribution|["{{PAExtendedHistogramContribution/bucket}}"].
+1. If |bucket| is a {{PASignalValue}}:
+    1. If |bucket|["{{PASignalValue/baseValue}}"] is not a valid [=signal base
+        value=], throw a {{TypeError}}.
+    1. If |bucket|["{{PASignalValue/offset}}"] is a {{long}}, throw a
```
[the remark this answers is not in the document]
alexmturner: Yep, agreed. Changed

spec.bs
alexmturner: I think these need to be bound to an auction as a whole rather than just a global scope as the spec doesn't keep track of which global scopes have been used in the past, so when we 'fill in' these contributions later, the global scopes might not be around any more.

But, yes, I think you're right that we should be exposing these values on the global scope itself so we don't need to use this handwavy prose. I've added an issue and plan to follow up in a separate PR shortly if that's ok with you.

spec.bs
alexmturner: See comment below, but yeah that makes sense. Will follow up

PR #66, opened 2023-06-15
Spec: Permissions policy integration

alexmturner:
Defines the "private-aggregation" permissions policy and plumbs values through. Improves other algorithm's plumbing as a drive-by.

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/66.html" title="Last updated on Jun 20, 2023, 7:52 PM UTC (e1e9da0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/66/270f49e...e1e9da0.html" title="Last updated on Jun 20, 2023, 7:52 PM UTC (e1e9da0)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -931,10 +931,11 @@ as appropriate:
     1. [=Assert=]: |globalScopes|' [=list/size=] equals 1.
     1. Let |privateAggregationObj| be |globalScopes|[0]'s
         {{SharedStorageWorkletGlobalScope/privateAggregation}}.
-    1. If [=this=]'s [=relevant global object=]'s [=associated document=] is
-        [=allowed to use=] "<code>[=private-aggregation=]</code>", set
-        |privateAggregationObj|'s [=PrivateAggregation/allowed to use=] to true.
-    1. Set |privateAggregationObj|'s [=PrivateAggregation/scoping details=] a
+    1. Set |privateAggregationObj|'s [=PrivateAggregation/allowed to use=] to
+        the result of determining whether [=this=]'s [=relevant global
+        object=]'s [=associated document=] is [=allowed to use=] the
+        "<code>[=private-aggregation=]</code>" [=policy-controlled feature=].
+    1. Set |privateAggregationObj|'s [=PrivateAggregation/scoping details=] to a
             new [=scoping details=] with the items:
         : [=scoping details/get batching scope steps=]
         :: An algorithm that returns the [=batching scope=] that is scheduled to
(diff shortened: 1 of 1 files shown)


alexmturner:
cc @yoavweiss

spec.bs
```diff
@@ -916,9 +929,13 @@ as appropriate:
     [=upon rejection=] of |promise|, run the following steps:
     1. Let |globalScopes| be |this|'s [=Worklet/global scopes=].
     1. [=Assert=]: |globalScopes|' [=list/size=] equals 1.
-    1. Set |globalScopes|[0]'s {{SharedStorageWorkletGlobalScope/
-        privateAggregation}}'s [=PrivateAggregation/scoping details=] a new
-            [=scoping details=] with the items:
+    1. Let |privateAggregationObj| be |globalScopes|[0]'s
+        {{SharedStorageWorkletGlobalScope/privateAggregation}}.
+    1. If [=this=]'s [=relevant global object=]'s [=associated document=] is
```
linnan-github, line 934: Optional: We can probably just set this field to the permission policy check without checking the value.

spec.bs
```diff
@@ -916,9 +929,13 @@ as appropriate:
...
+        [=allowed to use=] "<code>[=private-aggregation=]</code>", set
+        |privateAggregationObj|'s [=PrivateAggregation/allowed to use=] to true.
+    1. Set |privateAggregationObj|'s [=PrivateAggregation/scoping details=] a
```
linnan-github, line 937:
```suggestion
    1. Set |privateAggregationObj|'s [=PrivateAggregation/scoping details=] to a
```

linnan-github, line 937: Shall we early return if it's not allowed to use?

spec.bs
```diff
@@ -1071,35 +1090,49 @@ nested scope), renumbering this and later steps as necessary.
...
+|auctionConfig| and an [=interest group=] or null |ig|.
+
+Second, we add the following steps after step 11 ("If
 <var ignore>evaluationStatus</var> is an [=ECMAScript/abrupt completion=]..."),
 renumbering later steps as appropriate:
-<div algorithm="protected-audience-evaluate-a-script-monkey-patch-1">
-12. Let |debugScope| be a new [=debug scope=].
-1. Set <var ignore>global</var>'s {{InterestGroupScriptRunnerGlobalScope/
+<div algorithm="protected-audience-evaluate-a-script-monkey-patch">
+12. If |auctionConfig|'s [=auction config/permissions policy state=]'s
```
linnan-github, line 1101: Similar to shared storage, probably can just set the field without checking the value, and early return?

spec.bs
alexmturner: The current (implemented) behavior is actually to check the 'is initial execution' complete first, although we could consider swapping.

spec.bs

spec.bs

spec.bs
alexmturner: Switched the setting, but left early return as above

spec.bs
alexmturner: Yeah, that's fair -- I've added an issue to track whether we want to swap the order. I'll follow up soon

linnan-github (approved):

alexmturner:
Thanks!

PR #74, opened 2023-06-21
Spec: Minor wording tweaks

alexmturner:
Makes wording more consistent both internally and with other specs.


alexmturner:
(lower priority) @linnan-github, could you PTAL? Thanks!
cc @yoavweiss

linnan-github (approved):

alexmturner:
Thanks!

PR #75, opened 2023-06-23
Spec: Guard privateAggregation object for permissions and availability

alexmturner:
Guards the privateAggregation object directly instead of guarding each API function. This aligns with the implementation and reduces duplicate logic. Also reworks the wording of the permissions policy integration as a drive-by.

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/75.html" title="Last updated on Jul 6, 2023, 6:09 PM UTC (58ce26f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/75/ddb61fd...58ce26f.html" title="Last updated on Jul 6, 2023, 6:09 PM UTC (58ce26f)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -216,9 +216,9 @@ global scope. Its [=getter steps=] should be set to the [=get the
 privateAggregation=] steps given [=this=].
 
 Each global scope should set the [=PrivateAggregation/allowed to use=] for the
-{{PrivateAggregation}} object it exposes based on whether the
-"<code>[=private-aggregation=]</code>" [=policy-controlled feature=] is enabled
-for a relevant [=document=].
+{{PrivateAggregation}} object it exposes based on whether a relevant
+[=document=] is [=allowed to use=] the "<code>[=private-aggregation=]</code>"
+[=policy-controlled feature=].
 
 Additionally, each global scope should set the [=PrivateAggregation/scoping
 details=] for the {{PrivateAggregation}} object it exposes to a non-null value.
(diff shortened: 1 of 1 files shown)


alexmturner:
cc @yoavweiss

spec.bs
```diff
@@ -216,8 +209,15 @@ Issue: Ensure errors are of an appropriate type, e.g. {{InvalidAccessError}} is
...
 
-To expose this API to a global scope, a [=read only=] [=attribute=] `privateAggregation`
-of type {{PrivateAggregation}} should be exposed on the global scope.
+To expose this API to a global scope, a [=read only=] [=attribute=]
+`privateAggregation` of type {{PrivateAggregation}} should be exposed on the
+global scope. Its [=getter steps=] should be set to the [=get the
+privateAggregation=] steps given [=this=].
+
+Each global scope should set the [=PrivateAggregation/allowed to use=] for the
+{{PrivateAggregation}} object it exposes based on whether the
```
linnan-github, line 218: Maybe can rephrase to "based on whether the relevant document is allowed to use the "private-aggregation" policy control feature" and link properly?

spec.bs
```diff
@@ -837,6 +854,9 @@ partial dictionary SharedStorageRunOperationMethodOptions {
 };
 </xmp>
 
+The {{SharedStorageWorkletGlobalScope/privateAggregation}} getter steps are to
```
linnan-github, line 857: Shall we link getter steps?

linnan-github: Actually do we have to explicitly define the getter steps given that we already mention it should be set to [=get the privateAggregation=]?

spec.bs
alexmturner: Had a go along these lines

spec.bs
alexmturner: Linked getter steps for the two monkeypatches. I think unfortunately we probably have to do this as I don't think the Private Aggregation spec 'controls' the Shared Storage and Protected Audience getter steps (at least after the monkey patches are incorporated properly). It's a little weird that we instruct how implementers should expose their objects, but I think it's probably the layering that makes the most sense.

linnan-github: Yeah that makes sense. It’s a little weird, but not too bad.

linnan-github (approved):

PR #84, opened 2023-07-12
Clarify the intended contribution bounds

alexmturner:
The current wording is a little confusing.


alexmturner:

menonasha (approved):

Issue #23, opened 2023-03-14
Contribution bounding value and scope


alexmturner:
Hi all,

We're seeking some feedback on the Private Aggregation API's [contribution budget](https://github.com/patcg-individual-drafts/private-aggregation-api#contribution-bounding-and-budgeting). We'd appreciate any thoughts on both the value of the numeric bound as well as the [scope](https://github.com/patcg-individual-drafts/private-aggregation-api#partition-choice) (currently per-origin per-day and separate for FLEDGE and Shared Storage).

In particular, one change we're considering is moving the scope from per-origin to per-[site](https://web.dev/same-site-same-origin/#site). This would mitigate abuse potential for cases of things like wildcard domains which are (arguably) easier to mint than domains to exceed privacy limits. (See more discussion [here](https://github.com/WICG/attribution-reporting-api/issues/661).)

Thanks!

alexmturner:
In addition to the change from per-origin to per-site, we're considering changing the time component of the contribution bound. Specifically, we're considering moving the existing contribution bound (max value sum of 2^16) to applying over a 10 minute window instead of a daily window. We hope this will allow more flexibility and simplify budget management. As a backstop to prevent worst-case leakage, we're considering a new larger daily bound, e.g. 2^20. We'd appreciate any feedback on this proposal!

xottabut:
Hi Alex (@alexmturner ), I have a few questions regarding budget in context of Private Aggregation API.
I am reading this document to understand the budget for Private Aggregation API [contribution budget](https://github.com/patcg-individual-drafts/private-aggregation-api#contribution-bounding-and-budgeting)

- Is L1 budget per (1) one event (i.e. ad impression) or (2) all events that happens during the period X (last 24 hours)? The documentation says: "_each user agent will limit the contribution that it could make to the output of a query. In the case of a histogram operation, the user agent could bound the L1 norm of the values, i.e. the sum of all the contributions across all buckets_". It is also not completely clear, what is the query here.
[...]

PR #85, opened 2023-07-13
Improving description of contribution limits

alexmturner:
Clarify that limits are inherently a rate, not an absolute number. Addresses comments raised in #23.


alexmturner:

menonasha (approved):

PR #87, opened 2023-08-15
Spec: User agent abort flows

alexmturner:
This handles user settings and enrollment, among other reasons the user agent might want to drop reports.


alexmturner:

linnan-github (approved):

alexmturner:
(forgot to cc @yoavweiss)

PR #90, opened 2023-08-22
Spec: Allow debug mode to be disabled arbitrarily by the user agent.

alexmturner:
Adds logic that allows for debug mode to be disabled even if enableDebugMode() is called. This addresses issue #57. See also the corresponding explainer change: #93.

As a drive-by, adds an issue for an assert that is currently hit.

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/90.html" title="Last updated on Sep 21, 2023, 6:32 PM UTC (8f3888c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/90/cf92221...8f3888c.html" title="Last updated on Sep 21, 2023, 6:32 PM UTC (8f3888c)">Diff</a>

Issue #57, opened 2023-06-09
Consider tying `enableDebugMode()` availability to third-party cookie eligibility


alexmturner:
This would ensure parity with what is available through cookies and allow for the temporary debugging mechanism to 'automatically' be removed when third-party cookies are.

alexmturner:
Closing as this has been implemented

nikijohnson4:
Thanks for the update

a2425rdl:
good

PR #93, opened 2023-09-11
Clarify when debug mode is enabled

alexmturner:
Per issue #57, we tie debug mode to third-party cookie eligibility. See also the related spec change in #90.


alexmturner:
@linnan-github @menonasha, could you PTAL? Thanks!

menonasha (approved):

PR #94, opened 2023-09-11
Still send deterministic report when budget exhausted

alexmturner:
When a context ID is set, a report should be sent even if budget is exhausted to avoid leaking information about how much budget has been used.

See also the spec change: #91.


alexmturner:
@linnan-github @menonasha, could you PTAL? Thanks!

menonasha (approved):

PR #100, opened 2023-09-15
Detail reduced delay plans for context ID reports

### LINKED ISSUE #80: Consider reducing the delay for reports with a context ID
It may be possible to reduce the delay of reports sent with a context ID, see [here](https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/report_verification.md#reduced-delay).

While this change is web visible, the existing randomized delays mean that reliance on reports arriving at particular times is not currently possible. So, this should not be breaking.

alexmturner:
Updates the report verification explainer to describe our plans to reduce delays for reports with a context ID (instead of describing it as a hypothetical).

This should resolve #80.


alexmturner:
@menonasha @linnan-github, could you PTAL? Thanks!

linnan-github (approved):

menonasha (approved):

PR #108, opened 2023-12-14
Truncate not round scaled signal value

alexmturner:
This matches the behavior of the current implementation


alexmturner:

linnan-github (approved):

PR #113, opened 2024-01-26
Add link to current aggregation_coordinator_origin allowlist

rushilw-google:
I added the following to line 453 related to the recent PAA GCP update. PTAL and approve: 
See [allowlist](https://github.com/WICG/attribution-reporting-api/blob/main/aggregation_coordinator_origin_allowlist.md) of the aggregation service coordinator origins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,7 +414,8 @@ service deployed on AWS, GCP, and other platforms in the future. The specified
 origin would need to be on an allowlist maintained by the browser. If none is
 specified, a default will be used.
 
-This allowlist matches the Attribution Reporting API's, available [here](https://github.com/WICG/attribution-reporting-api/blob/main/aggregation_coordinator_origin_allowlist.md).
+This allowlist matches the Attribution Reporting API's, available 
+[here](https://github.com/WICG/attribution-reporting-api/blob/main/aggregation_coordinator_origin_allowlist.md).
 
 Shared Storage callers would specify this field when calling the `run()` or
 `selectURL()` APIs, e.g.
--- a/README.md
+++ b/README.md
@@ -454,6 +454,9 @@ This setting would be able to be overridden via the typical interest group
 mechanisms. For example, the update mechanism could support a new
 `privateAggregationConfig` key matching the call to `joinAdInterestGroup()`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -454,6 +454,9 @@ This setting would be able to be overridden via the typical interest group
 mechanisms. For example, the update mechanism could support a new
 `privateAggregationConfig` key matching the call to `joinAdInterestGroup()`.
 
+See [allowlist](https://github.com/WICG/attribution-reporting-api/blob/main/aggregation_coordinator_origin_allowlist.md)
```
alexmturner, line 457: i think it might make more sense to move this up just after line 415. We could also consider rewording to something like: "This allowlist matches the Attribution Reporting API's, available [here](URL)." Wdyt?

rushilw-google: That makes better sense - I made a new commit with the update, PTAL and see if that looks fine? 

Some of the language in the section around line 415 is "we propose", "we should", etc. If we've gone ahead with all these, should we update the language?

README.md
```diff
@@ -414,6 +414,8 @@ service deployed on AWS, GCP, and other platforms in the future. The specified
 origin would need to be on an allowlist maintained by the browser. If none is
 specified, a default will be used.
 
+This allowlist matches the Attribution Reporting API's, available [here](https://github.com/WICG/attribution-reporting-api/blob/main/aggregation_coordinator_origin_allowlist.md).
```
alexmturner, line 417: very minor: could we add a new line between `available` and `[here](...)`? (This won't affect the rendered document, but makes the plaintext a bit easier to read (with shorter lines)

README.md
alexmturner: Thanks, LGTM with one more nit.

Agreed that the language is a bit odd -- the explainer is meant for the "early/rough proposal" stage and isn't necessarily meant to be kept up-to-date like we've been doing. I think it's ok to leave as is given that, at least for now -- we're also trying to move our documentation to a more appropriate place.

README.md
rushilw-google: Done, thanks for the guidance!

alexmturner: Great, thanks!

Issue #92, opened 2023-09-08
Consider adding a custom 'label' to allow more flexible batching


alexmturner:
Currently, the aggregation service only allows each '[shared ID](https://github.com/WICG/attribution-reporting-api/blob/main/AGGREGATION_SERVICE_TEE.md#disjoint-batches)' to be present in one query. A set of reports with the same shared ID cannot be split for separate queries, even if the resulting batches are disjoint.

One option to add more flexibility is to support an optional, custom field (a ‘label’) that is factored into the shared ID generation. We could consider a few different options:

1. Putting the field in the shared_info: The reporting origin would be able to easily split reports into separate batches based on the label. However, this approach would require the label to be set outside the isolated (Shared Storage or Protected Audience) context. It also would require the report to be deterministic similar to the [context ID](https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/report_verification.md#deterministic-number-of-reports), i.e. sending a null report if no contributions are made. This approach is therefore unlikely to work for Protected Audience bidders (see related [discussion](https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/report_verification.md#specifying-a-contextual-id-and-each-possible-ig-owner)) and could increase the number of reports sent.
1. Putting the field in the payload: This avoids the deterministic report requirement and would allow the label to be based on cross-site data, i.e. set from inside the isolated contexts. But, this also prevents the reporting origin from directly determining the label embedded in the report. The reporting origin may therefore have to send a larger number of reports to the aggregation service and ask it to filter based on a given set of labels. For certain use cases, the reporting origin may be able to maintain a context ID to label mapping that would avoid this increased scale, albeit less ergonomically than above.
1. Allowing bucket range filtering: Instead of using an explicit label, we could allow filtering based on a range of buckets, with budget only used for that range. This could be more flexible but also increases the complexity of the Aggregation Service’s privacy budgeting implementation.
1. A combination of the above: We could implement multiple of the above options and allow them to be used together or in different situations.

For all of the above approaches, we’ll also need a mechanism to limit the scale impact on the Privacy Budget Service. For example, we want to prevent developers from specifying a unique ‘label’ per report. There are a few options we could consider, including:
1. The Aggregation Service could limit the number of labels/bucket ranges or shared IDs per query
1. We could limit the space of allowed labels/bucket ranges directly, e.g. only allowing integer labels up to a maximum value.

[...]

PR #123, opened 2024-03-29
Spec: filtering IDs

alexmturner:
Specs the ability to set a filtering ID (and modify the default ID space). See https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/flexible_filtering.md#proposal-filtering-id-in-the-encrypted-payload and issue #92.

To support this new functionality, we increase the report version. Note that this also requires aggregation service versions to support the new version.

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/123.html" title="Last updated on May 2, 2024, 4:29 PM UTC (3df2e53)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/123/e03bedb...3df2e53.html" title="Last updated on May 2, 2024, 4:29 PM UTC (3df2e53)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -171,14 +171,14 @@ are:
 1. Let |scopingDetails| be [=this=]'s [=PrivateAggregation/scoping details=].
 1. Let |batchingScope| be the result of running |scopingDetails|' [=scoping
     details/get batching scope steps=].
-1. Let |filteringIdByteSize| be the [=default filtering ID byte size=].
+1. Let |filteringIdMaxBytes| be the [=default filtering ID max bytes=].
 1. If [=pre-specified report parameters map=][|batchingScope|] [=map/exists=]:
-    1. Set |filteringIdByteSize| to [=pre-specified report parameters
+    1. Set |filteringIdMaxBytes| to [=pre-specified report parameters
         map=][|batchingScope|]'s [=pre-specified report parameters/filtering ID
-        byte size=].
+        max bytes=].
 1. If |contribution|["{{PAHistogramContribution/filteringId}}"] is not [=set/
     contained=] in [=the exclusive range|the range=] 0 to
-    256<sup>|filteringIdByteSize|</sup>, exclusive, [=exception/throw=] a
+    256<sup>|filteringIdMaxBytes|</sup>, exclusive, [=exception/throw=] a
     {{RangeError}}.
 1. Let |entry| be a new [=contribution cache entry=] with the items:
     : [=contribution cache entry/contribution=]
@@ -360,7 +360,7 @@ An aggregatable report is a [=struct=] with the following items:
 :: An [=aggregation coordinator=]
 : <dfn>context ID</dfn>
 :: A [=string=] or null
-: <dfn>filtering ID byte size</dfn>
+: <dfn>filtering ID max bytes</dfn>
 :: A positive integer
 : <dfn>queued</dfn>
 :: A [=boolean=]
@@ -392,7 +392,7 @@ items:
 <dl dfn-for="pre-specified report parameters">
 : <dfn>context ID</dfn> (default: null)
 :: A [=string=] or null
-: <dfn>filtering ID byte size</dfn> (default: [=default filtering ID byte size=])
+: <dfn>filtering ID max bytes</dfn> (default: [=default filtering ID max bytes=])
 :: A positive integer
 
 </dl>
@@ -431,12 +431,12 @@ parameters map=].
 Constants {#constants}
 ======================
 
-<dfn>Default filtering ID byte size</dfn> is a positive integer controlling the
-byte size used if none is explicitly chosen. Its value is 1.
+<dfn>Default filtering ID max bytes</dfn> is a positive integer controlling the
+max bytes used if none is explicitly chosen. Its value is 1.
 
-<dfn>Valid filtering ID byte size range</dfn> is a [=set=] of positive integers
-controlling the allowable byte sizes. Its value is [=the inclusive range|the
-range=] 1 to 8, inclusive.
+<dfn>Valid filtering ID max bytes range</dfn> is a [=set=] of positive integers
+controlling the allowable values of max bytes. Its value is [=the inclusive
+range|the range=] 1 to 8, inclusive.
 
 Issue: Consider adding more constants.
 
@@ -537,11 +537,11 @@ To <dfn>determine if a report should be sent deterministically</dfn> given a
 steps. They return a [=boolean=]:
 1. If |preSpecifiedParams|' [=pre-specified report parameters/context ID=] is
     not null, return true.
-1. If |preSpecifiedParams|' [=pre-specified report parameters/filtering ID byte
-    size=] is not the [=default filtering ID byte size=], return true.
+1. If |preSpecifiedParams|' [=pre-specified report parameters/filtering ID max
+    bytes=] is not the [=default filtering ID max bytes=], return true.
 1. Return false.
 
-Note: If a context ID or non-default filtering ID byte size was specified, a
+Note: If a context ID or non-default filtering ID max bytes was specified, a
     report is sent, even if there are no contributions or there is insufficent
     budget for the requested contributions. See [Protecting against leaks via
     the number of reports](#protecting-against-leaks-via-the-number-of-reports).
@@ -627,10 +627,10 @@ scope</dfn> given a [=pre-specified report parameters=] |params| and a
 1. Let |contextId| be |params|' [=pre-specified report parameters/context ID=].
 1. [=Assert=]: |contextId| is null or |contextId|'s [=string/length=] is not
     larger than 64.
-1. Let |filteringIdByteSize| be |params|' [=pre-specified report parameters/
-    filtering ID byte size=].
-1. [=Assert=]: |filteringIdByteSize| is [=set/contained=] in the [=valid
-    filtering ID byte size range=]
+1. Let |filteringIdMaxBytes| be |params|' [=pre-specified report parameters/
+    filtering ID max bytes=].
+1. [=Assert=]: |filteringIdMaxBytes| is [=set/contained=] in the [=valid
+    filtering ID max bytes range=]
 1. [=map/Set=] [=pre-specified report parameters map=][|batchingScope|] to
     |params|.
 
@@ -714,9 +714,9 @@ perform the following steps. They return an [=aggregatable report=].
     :: |aggregationCoordinator|
     : [=aggregatable report/context ID=]
     :: |preSpecifiedParams|' [=pre-specified report parameters/context ID=]
-    : [=aggregatable report/filtering ID byte size=]
-    :: |preSpecifiedParams|' [=pre-specified report parameters/filtering ID byte
-        size=]
+    : [=aggregatable report/filtering ID max bytes=]
+    :: |preSpecifiedParams|' [=pre-specified report parameters/filtering ID max
+        bytes=]
     : [=aggregatable report/queued=]
     :: false
 1. Return |report|.
@@ -954,11 +954,11 @@ To <dfn>obtain the plaintext payload</dfn> given an [=aggregatable report=]
         [below](#protecting-against-leaks-via-payload-size).
 1. [=list/iterate|For each=] |contribution| of |report|'s [=aggregatable report/
     contributions=]:
-    1. Let |filteringIdByteSize| be |report|'s [=aggregatable report/filtering
-        id byte size=].
+    1. Let |filteringIdMaxBytes| be |report|'s [=aggregatable report/filtering
+        id max bytes=].
     1. [=Assert=]: |contribution|["{{PAHistogramContribution/filteringId}}"]
         is [=set/contained=] in [=the exclusive range|the range=] 0 to
-        256<sup>|filteringIdByteSize|</sup>, exclusive.
+        256<sup>|filteringIdMaxBytes|</sup>, exclusive.
     1. Let |contributionData| be an [=ordered map=] of the following key/value
         pairs:
         : "`bucket`"
@@ -970,7 +970,7 @@ To <dfn>obtain the plaintext payload</dfn> given an [=aggregatable report=]
         : "`id`"
         :: The result of [=encoding an integer for the payload=] given
             |contribution|[="{{PAHistogramContribution/filteringId}}"] and
-            |filteringIdByteSize|.
+            |filteringIdMaxBytes|.
     1. [=list/Append=] |contributionData| to |payloadData|.
 1. Let |payload| be an [=ordered map=] of the following key/value pairs:
     : "`data`"
@@ -1085,7 +1085,7 @@ partial interface SharedStorageWorkletGlobalScope {
 dictionary SharedStoragePrivateAggregationConfig {
   USVString aggregationCoordinatorOrigin;
   USVString contextId;
-  [EnforceRange] unsigned long long filteringIdByteSize;
+  [EnforceRange] unsigned long long filteringIdMaxBytes;
 };
 
 partial dictionary SharedStorageRunOperationMethodOptions {
@@ -1134,17 +1134,17 @@ steps. They return a [=pre-specified report parameters=], null, or a
     |privateAggregationConfig|["{{SharedStoragePrivateAggregationConfig/contextId}}"].
 1. If |contextId|'s [=string/length=] is greater than 64, return a new
     {{DOMException}} with name "`DataError`".
-1. Let |filteringIdByteSize| be the [=default filtering ID byte size=].
-1. If |privateAggregationConfig|["{{SharedStoragePrivateAggregationConfig/filteringIdByteSize}}"]
-    [=map/exists=], set |filteringIdByteSize| to
-    |privateAggregationConfig|["{{SharedStoragePrivateAggregationConfig/filteringIdByteSize}}"].
-1. If |filteringIdByteSize| is not [=set/contained=] in the [=valid filtering ID
-    byte size range=], return a new {{DOMException}} with name "`DataError`".
+1. Let |filteringIdMaxBytes| be the [=default filtering ID max bytes=].
+1. If |privateAggregationConfig|["{{SharedStoragePrivateAggregationConfig/filteringIdMaxBytes}}"]
+    [=map/exists=], set |filteringIdMaxBytes| to
+    |privateAggregationConfig|["{{SharedStoragePrivateAggregationConfig/filteringIdMaxBytes}}"].
+1. If |filteringIdMaxBytes| is not [=set/contained=] in the [=valid filtering ID
+    max bytes range=], return a new {{DOMException}} with name "`DataError`".
 1. Return a new [=pre-specified report parameters=] with the items:
     : [=pre-specified report parameters/context ID=]
     :: |contextId|
-    : [=pre-specified report parameters/filtering ID byte size=]
-    :: |filteringIdByteSize|
+    : [=pre-specified report parameters/filtering ID max bytes=]
+    :: |filteringIdMaxBytes|
 
 The {{WindowSharedStorage}}'s {{WindowSharedStorage/run()}} method steps are
 modified in four ways. First, add the following steps just after step 2 ("If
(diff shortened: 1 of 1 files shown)


alexmturner:

spec.bs
```diff
@@ -496,6 +522,20 @@ scope=] |debugScope| and an optional [=debug details=] or null
         set |entry|'s [=contribution cache entry/debug details=] to
         |debugDetails|.
 
+To <dfn>determine if a null report should be sent</dfn> given a [=pre-specified
+report parameters=] |preSpecifiedParams|, perform the following steps. They
+return a [=boolean=]:
+1. If |preSpecifiedParams|' [=pre-specified report parameters/context ID=] is
+    not null, return true.
+1. If |preSpecifiedParams|' [=pre-specified report parameters/filtering ID byte
+    size=] is not 1, return true.
```
linnan-github, line 531: Shall we define an implemention-defined value for the default filtering ID byte size?

spec.bs
```diff
@@ -496,6 +522,20 @@ scope=] |debugScope| and an optional [=debug details=] or null
         set |entry|'s [=contribution cache entry/debug details=] to
         |debugDetails|.
 
+To <dfn>determine if a null report should be sent</dfn> given a [=pre-specified
```
linnan-github, line 525: Shall we rename the algorithm to something like `determine if a report should be sent deterministically` as it' not necessarily a null report?

spec.bs
```diff
@@ -573,12 +612,18 @@ Issue: Elsewhere, surround algorithms in a `<div algorithm>` block to match, and
...
 
-1. [=Assert=]: |contextId|'s [=string/length=] is not larger than 64.
-1. Let |contextIdMap| be the [=context ID map=].
-1. [=map/Set=] |contextIdMap|[|batchingScope|] to |contextId|.
+1. Let |contextId| be |params|' [=pre-specified report parameters/context ID=].
+1. [=Assert=]: |contextId| is null or |contextId|'s [=string/length=] is not
+    larger than 64.
+1. Let |filteringIdByteSize| be |params|' [=pre-specified report parameters/
+    filtering ID byte size=].
+1. [=Assert=]: |filteringIdByteSize| is an integer in the range [1, 8].
```
linnan-github, line 624: Probably consider defining implementation-defined values for the valid range of filtering ID byte size.

spec.bs
```diff
@@ -804,8 +850,26 @@ Note: This section is largely copied from the
...
+1. If |report|'s [=aggregatable report/debug details=]'s [=debug details/
+    enabled=] is true:
+    1. [=list/For each=] |contribution| of |report|'s [=aggregatable report/
+        contributions=]:
+        1. If |contribution|["{{PAHistogramContribution/filteringId}}"] [=map/
+            exists=]:
+            1. Set |shouldUseNewReportVersion| to true.
+            1. [=iteration/Break=].
+
+    Note: For now, we only upgrade a report to version 1.0 if debug mode is
```
linnan-github, line 864:
```suggestion
    Note: For now, we only upgrade a report to version "`1.0`" if debug mode is
```

spec.bs
```diff
@@ -804,8 +850,26 @@ Note: This section is largely copied from the
...
+        enabled and a filteringId was specified. This avoids affecting existing
+        flows while there are current Aggregation Service releases that do not
+        support reports with version 1.0. We additionally require debug mode so
```
linnan-github, line 867:
```suggestion
        support reports with version "`1.0`". We additionally require debug mode so
```

spec.bs
```diff
@@ -804,8 +850,26 @@ Note: This section is largely copied from the
...
+        that the report version cannot be used to leak any new information. Once
+        all current Aggregation Service releases support version 1.0, we can
```
linnan-github, line 869:
```suggestion
        all current Aggregation Service releases support version "`1.0`", we can
```

spec.bs
```diff
@@ -826,17 +891,17 @@ error.
     value to JSON bytes=] on |data|.
 
 To <dfn>obtain the aggregation service payloads</dfn> given an [=aggregatable
-report=] |report|, perform the following steps. They return a [=list=] of
-[=maps=] or an error.
+report=] |report| and [=boolean=] |shouldUseNewReportVersion|, perform the
```
linnan-github, line 894:
```suggestion
report=] |report| and a [=boolean=] |shouldUseNewReportVersion|, perform the
```

spec.bs
```diff
@@ -877,7 +942,8 @@ Note: The user agent is encouraged to enforce regular key rotation. If there are
     random for every encryption operation.
 
 To <dfn>obtain the plaintext payload</dfn> given an [=aggregatable report=]
-    |report|, perform the following steps. They return a [=byte sequence=].
+    |report| and [=boolean=] |shouldUseNewReportVersion|, perform the following
```
linnan-github, line 945:
```suggestion
    |report| and a [=boolean=] |shouldUseNewReportVersion|, perform the following
```

spec.bs
```diff
@@ -939,7 +1024,8 @@ big-endian [=byte sequence=] of length |bitLength| / 8, left padding with zeroes
 as necessary.
 
 To <dfn>obtain a report's shared info</dfn> given an [=aggregatable report=]
-|report|, perform the following steps. They return a [=string=].
+|report| and [=boolean=] |shouldUseNewReportVersion|, perform the following
```
linnan-github, line 1027:
```suggestion
|report| and a [=boolean=] |shouldUseNewReportVersion|, perform the following
```

linnan-github, line 1027:
```suggestion
|report| and a [=boolean=] |shouldUseNewReportVersion|, perform the following
```

spec.bs
```diff
@@ -168,12 +169,21 @@ are:
...
     [=exception/throw=] a {{RangeError}}.
 1. Let |scopingDetails| be [=this=]'s [=PrivateAggregation/scoping details=].
+1. Let |batchingScope| be the result of running |scopingDetails|' [=scoping
+    details/get batching scope steps=].
+1. Let |filteringIdByteSize| be 1.
+1. Let |preSpecifiedParamsMap| be the [=pre-specified report parameters map=].
+1. If |preSpecifiedParamsMap|[|batchingScope|] [=map/exists=]:
+    1. Set |filteringIdByteSize| to |preSpecifiedParamsMap|[|batchingScope|]'s
+        [=pre-specified report parameters/filtering ID byte size=].
+1. If |contribution|["{{PAHistogramContribution/filteringId}}"] is not in the
```
linnan-github, line 179: `filteringId` is optional, shall we check whether it's null first?

alexmturner: Ah good catch, done!

spec.bs
alexmturner: Done!

spec.bs

spec.bs
```diff
@@ -372,6 +384,19 @@ A context type is a [=string=] indicating what kind of global scope the
...
+Pre-specified report parameters {#pre-specified-report-parameters-structure}
+----------------------------------------------------------------------------
+
+A <dfn>pre-specified report parameters</dfn> is a [=struct=] with the following
+items:
+<dl dfn-for="pre-specified report parameters">
+: <dfn>context ID</dfn> (default null)
+:: A [=string=] or null
+: <dfn>filtering ID byte size</dfn> (default 1)
+:: A positive integer.
```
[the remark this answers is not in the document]

spec.bs

spec.bs

spec.bs

spec.bs
alexmturner: Makes sense! Done

spec.bs
alexmturner: As above

alexmturner:
Thanks for the review, Nan!

alexmturner:
@arichiv, could you PTAL as well? Thanks!

alexmturner:
Two additional things I'm seeking feedback on:
- I'm a little unsure what to name `filteringIdByteSize`. For context, this allows for using a custom size for the filtering ID space, e.g. you could specify 4 to allow for a 32-bit ID. (We only allow integer bytes which is why we went with ByteSize not BitSize.) Other options we were thinking of is `filteringIdByteLength` or `filteringIdMaxBytes` (maybe leaning towards this).
- The default `filteringIdByteSize` is 1. When a non-default value is used, a report is sent even if no contributions were sent (a "null report"). This is done to avoid a privacy concern from the different payload length. However, it might be easier to describe the behavior if we activate this null reporting logic whenever a `filteringIdByteSize` is explicitly specified, even if it's for the default value. Personally, I'm probably leaning against this as it's not needed for privacy.

spec.bs
linnan-github: Even if it’s not implementation defined, maybe we can follow the ARA spec which defines constant values for readability?

alexmturner: Yep, added!

alexmturner:
@linnan-github @arichiv, I've updated this PR to no longer have an explicit opt-in or require debug mode. This should simplify the change substantially, albeit potentially delay the launch slightly (as we will now want all current aggregation service releases to support the new format before launching to Chrome Stable).

spec.bs
```diff
@@ -2311,10 +2422,10 @@ Security considerations {#security-considerations}
 Same-origin policy {#same-origin-policy}
 ----------------------------------------
 
-Writes to the [=aggregatable report cache=], [=context ID map=], [=contribution
-cache=] and [=debug scope map=] are separated by the reporting [=origin=] and
-the data included in any report with a given reporting [=origin=] are generated
-with only data from that [=origin=].
+Writes to the [=aggregatable report cache=], [=contribution cache=], [=debug
```
arichiv, line 2425: I'm not sure how much I like this statement, it's true every report has an origin, each cache entry has a debug and batching scope, and the other two are mapped to unique scopes.

When I dig into the details it does seem that the underlying scopes or operations are attributed to a given origin, but I'm not sure they are separated by origin exactly. As the current storage model stands, one has the information needed to separate them by origin but the lists at least are not really separated at all. 

Anyway, this is a long way to say I think 'attributed to' is a bit more correct than 'separated by'.

arichiv (changes requested):

arichiv:
It might be worth splitting this change into two parts:
1. replacing `context id` with a `pre-specified report parameters` struct that contains only a `context id`
2. adding a `filtering ID byte size` to the `pre-specified report parameters` struct

Doing both at once when so many spots are touched by 1 but only a few feel the impact of 2 is risky from a comprehensive review perspective.

spec.bs
```diff
@@ -517,19 +568,18 @@ a [=batching scope=] |batchingScope|, an [=origin=] |reportingOrigin|, a
...
         |aggregationCoordinatorMap|[|batchingScope|].
     1. [=map/Remove=] |aggregationCoordinatorMap|[|batchingScope|].
-1. Let |contextId| be null.
-1. Let |contextIdMap| be the [=context ID map=].
-1. If |contextIdMap|[|batchingScope|] [=map/exists=]:
-    1. Set |contextId| to |contextIdMap|[|batchingScope|].
-    1. [=map/Remove=] |contextIdMap|[|batchingScope|].
-1. Otherwise, [=assert=]: |timeout| is null.
+1. Let |preSpecifiedParams| be a new [=pre-specified report parameters=].
+1. Let |preSpecifiedParamsMap| be the [=pre-specified report parameters map=].
```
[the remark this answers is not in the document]
alexmturner: Agreed that this is a bit awkward -- the reason I went with these local variables is for the indexing syntax as it doesn't quite work for multi-word references given the spaces, e.g. `[=pre-specified report parameters map=][|batchingScope|]`. What would you recommend for this? We could spell it out with words, but I'm a little worried it hurts readability, e.g. something like:
```
1. If the [=pre-specified report parameters map=] [=map/contains an entry with the given key=] |batchingScope|:
    1. Set |filteringIdByteSize| to the [=pre-specified report parameters/filtering ID byte size=] of the result of [=map/getting the value of an entry=] in the [=pre-specified report parameters map=] given |batchingScope|.
```

I suppose I could also consider parentheses or something similar -- not sure what's common for specs.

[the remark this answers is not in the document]
arichiv: Maybe i'm misunderstanding, I think it's possible to do  [=...=][|...|], I do it here: https://github.com/w3c/webappsec-permissions-policy/blob/7a2bc78e99c7cf3f268c9932ca02e3024238049d/index.bs#L701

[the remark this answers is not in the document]
alexmturner: Ack, thanks. I think the indexing notation with a multi-word variable just looked a bit wrong to my eyes, but good to know it's used in other specs. I've put up a PR (#127) to clean up some of the existing usage in the spec.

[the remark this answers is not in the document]
alexmturner: Fixed this instance in #128

spec.bs
```diff
@@ -1111,12 +1198,15 @@ renumbering the last step as appropriate):
         |batchingScope|, <var ignore>outsideSettings</var>' [=environment
         settings object/origin=], "<code>shared-storage</code>" and
         |privateAggregationTimeout|.
-1. If |contextId| is not null:
-    1. Set |privateAggregationTimeout| to the [=current wall time=] plus a
-        non-negative [=implementation-defined=] [=duration=].
-    1. [=Set the context ID for a batching scope=] given |contextId| and
-        |batchingScope|.
-    1. Run the following steps [=in parallel=]:
+1. If |preSpecifiedParams| is not null:
```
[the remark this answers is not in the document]
alexmturner: This one was a mistake -- we shouldn't be repeating this. Fixed this in #128.

spec.bs
```diff
@@ -1162,12 +1252,15 @@ be |operationMap|[|name|]." (renumbering later steps as appropriate):
         and |privateAggregationTimeout|.
 1. If |aggregationCoordinator| is not null, [=set the aggregation coordinator
     for a batching scope=] given |aggregationCoordinator| and |batchingScope|.
-1. If |contextId| is not null:
-    1. Set |privateAggregationTimeout| to the [=current wall time=] plus a
-        non-negative [=implementation-defined=] [=duration=].
-    1. [=Set the context ID for a batching scope=] given |contextId| and
-        |batchingScope|.
-    1. Run the following steps [=in parallel=]:
+1. If |preSpecifiedParams| is not null:
```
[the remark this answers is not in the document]
alexmturner: This one I think I might leave for now -- agreed that it's duplicate, but I don't think this PR changes that too much. I'm hoping to do a better job when we move this into the Shared Storage spec.

alexmturner (commented):
Thanks for the review! Pulled out the no-op pre-specified report parameter changes into their own PR (#128). I've left comments about other issues unresolved for now and will address them in this PR once its rebased onto the other one.

spec.bs
```diff
@@ -905,6 +969,14 @@ To <dfn>obtain the plaintext payload</dfn> given an [=aggregatable report=]
         : "`value`"
         :: The result of [=encoding an integer for the payload=] given
             |contribution|["{{PAHistogramContribution/value}}"] and 32.
+        : "`id`"
+        :: The result of [=encoding an integer for the payload=] given
+            |contribution|["{{PAHistogramContribution/filteringId}}"] and
+            |filteringIdBitSize|.
+
+            Issue: Be consistent about bytes or bits as units here.
```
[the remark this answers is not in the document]
alexmturner: Nope, resolved.

alexmturner (commented):
This is ready for review again. I think I've addressed the remaining comments and rebased appropriately.

spec.bs
```diff
@@ -1349,9 +1408,16 @@ event, PAExtendedHistogramContribution contribution)</dfn> method steps are:
         throw=] a {{TypeError}}.
 1. Otherwise, if |contribution|["{{PAHistogramContribution/value}}"] is
     negative, [=exception/throw=] a {{TypeError}}.
+1. If |contribution|["{{PAExtendedHistogramContribution/filteringId}}"] is
```
arichiv, line 1411: This is the third time we do something similar, it might be worth making an algo but up to you

arichiv (approved):
Note: I have not done a detailed review of the entire spec but the change just here look good to me. Thanks for splitting the PR into multiple parts, that made it a lot easier to read through

alexmturner:
Thanks for the review, @arichiv! I was also wondering if you had any feedback on the questions in https://github.com/patcg-individual-drafts/private-aggregation-api/pull/123#issuecomment-2070958555.

arichiv:
> Two additional things I'm seeking feedback on:
> 
> * I'm a little unsure what to name `filteringIdByteSize`. For context, this allows for using a custom size for the filtering ID space, e.g. you could specify 4 to allow for a 32-bit ID. (We only allow integer bytes which is why we went with ByteSize not BitSize.) Other options we were thinking of is `filteringIdByteLength` or `filteringIdMaxBytes` (maybe leaning towards this).

I like filteringIdMaxBytes given the usage I see.

> * The default `filteringIdByteSize` is 1. When a non-default value is used, a report is sent even if no contributions were sent (a "null report"). This is done to avoid a privacy concern from the different payload length. However, it might be easier to describe the behavior if we activate this null reporting logic whenever a `filteringIdByteSize` is explicitly specified, even if it's for the default value. Personally, I'm probably leaning against this as it's not needed for privacy.

Is there a design doc I could review? I'm not sure I understand the implications from what I've reviewed alone

alexmturner:
We have the explainer here: https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/flexible_filtering.md#small-id-space-by-default-but-configurable.

To elaborate on the privacy concern, when reports are _not_ deterministically sent, the count of reports is potentially sensitive, i.e. based on cross-site data. For example, a script could choose whether or not to make contributions in a shared storage worklet (and a report would only be sent if there are contributions). The filtering ID feature allows scripts to also affect payload size (although only based on first-party data), which could allow an adversary to "partition" its counts -- i.e. it could separately count the number of reports with each different payload size to gain more information. Thus, this additional metadata would amplify the existing counting attack.

We prevent this by ensuring that, whenever a non-default byte size is chosen (and so a non-default payload size will occur) reports are sent deterministically, i.e. even if the script makes no contributions. This means that the count of reports with that payload size can no longer be affected by cross-site data.

The question here is whether we should also extend this protection to cases where a script has explicitly chosen the default byte size. There is no privacy reason to do this -- these reports are indistinguishable from reports that got the default due to not specifying anything. However, it arguably might be easier for developers to grok the "use the param => deterministic report" logic over "specify a non-default value => deterministic report" logic. There is also a small performance downside to this -- a report will be needlessly sent if no contributions are made in this case. I'm leaning against this due to this small performance impact and because it feels more natural to align the logic with the privacy requirement; it also means we can define the WebIDL in the form `type filteringIdMaxBytes = 1` and not have special "not present" logic.

(And no worries if you don't really have opinions here -- happy to resolve internally if so)

arichiv:
I get it now, thanks for the elaboration. I would suggest something slightly more elaborate: modifying `SharedStoragePrivateAggregationConfig` to take both a `filteringIdMaxBytes` and a new enum called something like `reportingStrategy` which is a string enum that can be `wait-for-contributions` or `always-send` (open to other naming).

`filteringIdMaxBytes` has the same current properties (defaults to 1).

`reportingStrategy` defaults to `wait-for-contributions` when `filteringIdMaxBytes` is 1 and `always-send` otherwise (doesn't distinguish between `filteringIdMaxBytes` being 1 explicitly or by default). If `reportingStrategy` is set to `wait-for-contributions` when `filteringIdMaxBytes` isn't 1 an error is thrown. If `reportingStrategy`  is set to `always-send` when `filteringIdMaxBytes` is 1 that's fine.

arichiv (changes requested):
Based on last comment

alexmturner:
Just switched the spec over to using "max bytes".

Oh interesting! I do see the appeal of making this property a bit more explicit, but I don't think any developer will want to opt-in to additional null reports if they can avoid it. Chatted briefly internally and, given that, we were thinking that leaving the logic as is should be simple enough for developers (and aligns with this idea that they want to minimize null reports when possible).

That being said, we could consider adopting randomized null reporting for the non-deterministic case in the future, similar to [ARA's proposal](https://github.com/WICG/attribution-reporting-api/blob/main/AGGREGATE.md#hide-the-true-number-of-attribution-reports). If we do end up pursuing that, we might want to return to this idea as there would be more of a meaningful tradeoff.

Thanks again both of you for all the reviews!

arichiv (approved):
It sounds like if the default or an explicit 1 is used then empty reports won't be sent? I'm fine with that. I'm just against having defaults behave differently from explicit settings identical to the defaults.

alexmturner:
Yep, that's right!

spec.bs
```diff
@@ -1038,6 +1085,7 @@ partial interface SharedStorageWorkletGlobalScope {
 dictionary SharedStoragePrivateAggregationConfig {
   USVString aggregationCoordinatorOrigin;
   USVString contextId;
+  [EnforceRange] unsigned long long filteringIdMaxBytes;
```
linnan-github, line 1088: Is there a reason to use long long for max bytes?

arichiv: That's a good point, doesn't JS max out at 54 bits for ints because that's the non-exponent part of a double (and all JS numbers are doubles)? Is an unsigned long enough?

alexmturner: This is just following the guidance in the Web Platform Design Principles [here](https://www.w3.org/TR/design-principles/#numeric-types). We unfortunately didn't follow this for some of the other existing fields -- we may want to change those at some point.

linnan-github (approved):

PR #129, opened 2024-05-03
Minor updates to filtering ID explainer

alexmturner:
Updates the variable name (ByteSize -> MaxBytes), the limit on the ID space (from 32 bits/4 bytes to 64 bits/8 bytes) and to make the filteringId's type a bigint.


alexmturner:

PR #133, opened 2024-05-30
Align enableDebugMode() semantics in explainer with spec

dmcardle:
The spec defines [method steps for enableDebugMode()](https://patcg-individual-drafts.github.io/private-aggregation-api/#dom-privateaggregation-enabledebugmode). These steps indicate that the method should throw a `DataError` exception when it has already been called in the current scope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,6 +325,9 @@ privateAggregation.enableDebugMode();
 The browser can optionally apply debug mode to reports requested earlier in that
 context.
 
+This javascript function can only be called once per context. Any subsequent
+calls will throw an exception.
+
 #### Debug keys
 To allow sites to associate reports with the contexts that triggered them, we
 also allow setting 64-bit unsigned integer debug keys. These keys are passed as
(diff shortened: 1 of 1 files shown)


dmcardle:
@alexmturner PTAL when you're available!

johnivdel:
Could you link the part of the spec which captures this rather than the crbug?

dmcardle:
Done!

README.md
```diff
@@ -334,7 +337,7 @@ privateAggregation.enableDebugMode({debugKey: 1234n});
 ```
 
 This javascript function can only be called once per context. Any subsequent
```
alexmturner, line 339: Maybe let's just delete this repeated paragraph, putting the ref under "Enabling" probably makes more sense :)

dmcardle: Done!

alexmturner (approved):
LGTM, thanks!

dmcardle:
Thanks! This is ready to go, but I don't have write access :)

alexmturner:
Oh huh, not seeing the setting to change that right now, so I'll just go ahead and merge for now :)

PR #136, opened 2024-06-28
Spec: client-side contribution merging

alexmturner:
Adds support for merging contributions with the same bucket and filtering ID before truncating to the contribution limit.

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/136.html" title="Last updated on Jun 28, 2024, 6:20 PM UTC (6a21776)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/136/a321c7c...6a21776.html" title="Last updated on Jun 28, 2024, 6:20 PM UTC (6a21776)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -654,7 +654,7 @@ null |timeout|:
 1. Let |mergedContributions| be a new [=list=].
 1. [=list/For each=] |contribution| of |contributions|:
     1. Let |hasProcessedContribution| be false.
-    1. [=list/For each] |mergedContribution| of |mergedContributions|:
+    1. [=list/For each=] |mergedContribution| of |mergedContributions|:
         1. If |contribution| and |mergedContribution| have both the same
             {{PAHistogramContribution/bucket}} and the same
             {{PAHistogramContribution/filteringId}}:
--- a/spec.bs
+++ b/spec.bs
@@ -651,13 +651,27 @@ null |timeout|:
         agents=] to drop reports for a number of reasons, for example user
         opt-out or an origin not being
         <a href="https://github.com/privacysandbox/attestation">enrolled</a>.
+1. Let |mergedContributions| be a new [=list=].
+1. [=list/For each=] |contribution| of |contributions|:
+    1. Let |hasProcessedContribution| be false.
+    1. [=list/For each] |mergedContribution| of |mergedContributions|:
+        1. If |contribution| and |mergedContribution| have both the same
+            {{PAHistogramContribution/bucket}} and the same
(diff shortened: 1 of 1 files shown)


alexmturner:
@dmcardle could you PTAL? Thanks!

spec.bs
```diff
@@ -651,13 +651,27 @@ null |timeout|:
...
         <a href="https://github.com/privacysandbox/attestation">enrolled</a>.
+1. Let |mergedContributions| be a new [=list=].
+1. [=list/For each=] |contribution| of |contributions|:
+    1. Let |hasProcessedContribution| be false.
+    1. [=list/For each=] |mergedContribution| of |mergedContributions|:
+        1. If |contribution| and |mergedContribution| have both the same
+            {{PAHistogramContribution/bucket}} and the same
+            {{PAHistogramContribution/filteringId}}:
+            1. Add |contribution|'s {{PAHistogramContribution/value}} to
+                |mergedContribution|'s {{PAHistogramContribution/value}}.
```
dmcardle, line 662: Pedantic question: do the spec semantics literally mean that this step mutates `mergedContributions`? I read [list-iterate](https://infra.spec.whatwg.org/#list-iterate) but was still unsure.

spec.bs
```diff
@@ -651,13 +651,27 @@ null |timeout|:
         agents=] to drop reports for a number of reasons, for example user
         opt-out or an origin not being
         <a href="https://github.com/privacysandbox/attestation">enrolled</a>.
+1. Let |mergedContributions| be a new [=list=].
```
dmcardle, line 654: Nit: Is it possible to express `mergedContributions` as a map from (bucket, filteringId) to values? It would eliminate the need for `hasProcessedContribution`, but I suppose it wouldn't be much simpler for the reader.

dmcardle (approved):
LGTM! Sorry for the delay.

spec.bs
alexmturner: I think the infra spec doesn't currently elaborate on what happens with mutation during iteration, but my thinking was that this should be fine given that no elements are inserted or deleted.

spec.bs
alexmturner: I was trying to prioritize readability here, so i think I would prefer it without the map. (Note that any algorithm that has the same end result is considered [conformant](https://infra.spec.whatwg.org/#algorithm-conformance).)

PR #147, opened 2024-07-25
Spec: Fix algorithm declarations

dmcardle:
The Bikeshed Documentation (dated 8 March 2024) says, "Algorithms can be explicitly indicated in your markup by putting the algorithm attribute on a container element or a heading" [1].

Through trial and error, I discovered that algorithm-specific markup is not generated when you write `<dfn algorithm>`. Instead, the `<dfn>` tag must be wrapped in something like `<div algorithm>`.

This change has at least two positive results:

1. Local variables are no longer declared in the global scope!
2. We can now click on local variables and they are highlighted in pretty colors.

[1]: https://speced.github.io/bikeshed/#var-and-algorithms

***
<a href="https://pr-preview.s3.amazonaws.com/dmcardle/private-aggregation-api/pull/147.html" title="Last updated on Aug 19, 2024, 2:59 PM UTC (d47f727)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/147/94f3f55...dmcardle:d47f727.html" title="Last updated on Aug 19, 2024, 2:59 PM UTC (d47f727)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1073,10 +1073,9 @@ They return a [=byte sequence=] or an error.
     [[RFC9180#name-encryption-to-a-public-key|sender's context]] with |pkR|,
     |info|, |kem_id|, |kdf_id| and |aead_id|.
 1. Let |aad| be \`\` (an empty [=byte sequence=]).
-1. Let |encryptedPayload| be the result of
+1. Let <var ignore>encryptedPayload</var> be the result of
     [[RFC9180#name-encryption-and-decryption|encrypting]] |plaintextPayload|
     with |hpkeContext| and |aad|.
-1. Return the [=byte sequence=] |encryptedPayload|.
 
 </div>
 
--- a/spec.bs
+++ b/spec.bs
@@ -162,7 +162,6 @@ Note: See [Exposing to global scopes](#exposing) below.
 The <dfn method for="PrivateAggregation">
 contributeToHistogram(PAHistogramContribution contribution)</dfn> method steps
 are:
-</div>
 
 1. If |contribution|["{{PAHistogramContribution/bucket}}"] is not [=set/
     contained=] in [=the exclusive range|the range=] 0 to 2<sup>128</sup>,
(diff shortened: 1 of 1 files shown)


dmcardle:
@alexmturner PTAL!

I'm not sure how exactly the Shared Storage monkey patches should be marked up, but there's an `Issue` tracking that in spec.bs.

spec.bs
```diff
@@ -1206,7 +1279,7 @@ be |operationMap|[|name|]." (renumbering later steps as appropriate):
 </div>
 
 Third, add the following steps in the same nested scope just before the current
-penultimate step ("If |options| [=map/contains=] |data|", renumbering the last
+penultimate step ("If <var ignore=''>options</var> [=map/contains=] <var ignore=''>data</var>", renumbering the last
```
alexmturner, line 1282: nit: line wrap, also no need for the `=''` after `ignore` (here and elsewhere)

spec.bs
```diff
@@ -2089,15 +2164,17 @@ an <a spec="turtledove">auction config</a> |auctionConfig| and a
                 the [=contribution cache=].
 1. [=Mark a debug scope complete=] given |auctionReportBuyersDebugScope| and
     |auctionConfig|'s [=auction config/auction report buyer debug details=].
-1. [=map/For each=] (|origin|, |aggregationCoordinator|) → |batchingScope| of
+1. [=map/For each=] (|origin|, <var ignore=''>aggregationCoordinator</var>) → |batchingScope| of
```
alexmturner, line 2167: nit: wrap

alexmturner (commented):
Thanks, definitely an improvement! Mainly just needs a rebase (and hopefully that'll fix the presubmit too)

dmcardle:
@alexmturner PTAL!

spec.bs
```diff
@@ -1021,12 +1076,19 @@ They return a [=byte sequence=] or an error.
 1. Let |encryptedPayload| be the result of
     [[RFC9180#name-encryption-and-decryption|encrypting]] |plaintextPayload|
     with |hpkeContext| and |aad|.
+1. Return the [=byte sequence=] |encryptedPayload|.
```
alexmturner, line 1079: Oh hmm this algorithm does seem broken. Happy either way, but wondering if this would make more sense to do in a separate PR than as a drive-by. Separately, I think we need to handle the case that the hpke setup or encryption fails.

E.g. `If |hpkeContext| is an error, return |hpkeContext|`

dmcardle: I'd punt semantic changes to another PR :)

alexmturner: Sg -- let's maybe remove this line too then :)

dmcardle: Done! Just pushed a fixup commit that drops the new "return" step and adds the `ignore` attribute to the now unused `encryptedPayload`. I'll make a followup PR that attempts to handle failing HPKE operations and brings back the return.

dmcardle:
Hey Alex, PTAL :)

alexmturner (approved):
LGTM!

PR #149, opened 2024-07-26
[CI] Change preview.yml to validate.yml

dmcardle:
This change converts the "Preview" job into a "Validate" job. It still runs for every PR, but it does not deploy to the gh-pages branch.

Fixes #148

***
<a href="https://pr-preview.s3.amazonaws.com/dmcardle/private-aggregation-api/pull/149.html" title="Last updated on Jul 26, 2024, 3:35 PM UTC (7f5429c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/149/11c6711...dmcardle:7f5429c.html" title="Last updated on Jul 26, 2024, 3:35 PM UTC (7f5429c)">Diff</a>

Issue #81, opened 2023-06-28
Consider splitting contributions into multiple reports instead of truncating


alexmturner:
Feedback from discussion in #44. See also [this explainer section](https://github.com/patcg-individual-drafts/private-aggregation-api#contributions-limit). The interaction between this and a providing context ID needs to be thought through still. To avoid performance concerns, we will likely still need a limit on the number of reports.

This would be a web visible change as number of reports being sent would change. However, the main compatibility concern is around feature detection -- developers may wish to know if they can safely use more contributions. One option is to provide a method/attribute with the maximum number of allowed contributions.

csharrison:
I want to also think about this in the context of performance sensitive callers that want to make sure padding (https://github.com/patcg-individual-drafts/private-aggregation-api/issues/56) is not bloating storage. E.g. if `context_id`-providing callers are sure they will always make a single contribution, we should let them pre-specify that and receive a more optimized payload.

maplgh:
The truncation to 20 contributions proposed and introduced in [issue#32](https://github.com/patcg-individual-drafts/private-aggregation-api/issues/32) hinders our use of the Private Aggregation API (PAA).

We are considering using PAA to obtain debug and troubleshooting data and other use cases. The debug and troubleshooting metrics will require multiple auction scope contributions (e.g. auction count, win count, loss count) and at least one candidate scope contribution to record rejection reasons. Based on our analysis, just the debug and troubleshooting contributions will hit the limit of 20, and we still have the other use cases that we are considering.

Would be great to learn about the status of this issue about splitting contributions into multiple reports.

dmcardle:
To avoid complex privacy mitigations, we're no longer considering splitting contributions into multiple reports. Instead, we are designing a pair of solutions that enable sites to customize the number of contributions per report.

## Stopgap for Protected Audience callers

We recognize that 20 contributions are not enough for some use cases and that it will take some time to launch a more thorough solution. Until then, we’re considering increasing the default — for Protected Audience script runners only — from 20 contributions per report to 100.

Raising the limit would certainly improve utility, but it would also increase the costs for adtechs. Estimates show that processing reports with 100 contributions could roughly double the processing time and cost of operating the Aggregation Service. Changing the default would affect all adtechs, even those that may not need additional contributions.

We’re seeking feedback from adtechs: is this appealing as a temporary solution or is the cost too high?

## Solution for callers without a context ID

[...]

PR #150, opened 2024-07-26
Spec: Define per-calling-API limit on contributions per report

dmcardle:
This PR corresponds to
  * "*Increase default report size for Protected Audience callers*" in <https://github.com/patcg-individual-drafts/private-aggregation-api/issues/81#issuecomment-2091524214>.
  * The explainer change in PR #138.

~~Note that this PR depends on #149. (GitHub doesn't understand dependent PRs, so this appears to include its parent's changes.)~~

***
<a href="https://pr-preview.s3.amazonaws.com/dmcardle/private-aggregation-api/pull/150.html" title="Last updated on Sep 3, 2024, 4:14 PM UTC (5c54351)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/150/9108e9b...dmcardle:5c54351.html" title="Last updated on Sep 3, 2024, 4:14 PM UTC (5c54351)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -456,8 +456,10 @@ controls which [=origins=] are valid [=aggregation coordinators=]. Every
 <dfn>Default aggregation coordinator</dfn> is an [=aggregation coordinator=]
 that controls which is used for a report if none is explicitly selected.
 
-<dfn>Maximum report contributions</dfn> is a positive integer that controls how
-many contributions can be present in a single report.
+<dfn>Maximum report contributions</dfn> is a [=map=] from [=context type=] to
+positive integers. Semantically, it defines the maximum number of contributions
+that can be present in a single report for every kind of calling context, e.g.
+Shared Storage.
 
 <dfn>Minimum report delay</dfn> is a non-negative [=duration=] that controls the
 minimum delay to deliver an [=aggregatable report=].
@@ -690,9 +692,9 @@ null |timeout|:
         |mergedContributions|.
 1. Let |truncatedContributions| be a new [=list=].
 1. If |mergedContributions| has a [=list/size=] greater than [=maximum report
-    contributions=]:
+    contributions=][|api|]:
     1. [=set/For each=] |n| of [=the exclusive range|the range=] 0 to [=maximum
-        report contributions=], exclusive:
+        report contributions=][|api|], exclusive:
         1. [=set/Append=] |mergedContributions|[|n|] to
             |truncatedContributions|.
 1. Otherwise, set |truncatedContributions| to |mergedContributions|.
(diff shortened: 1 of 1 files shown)


dmcardle:
@alexmturner PTAL :)

(Yay, CI gave me a green checkmark because I rebased on #149.)

spec.bs
```diff
@@ -451,8 +451,10 @@ controls which [=origins=] are valid [=aggregation coordinators=]. Every
 <dfn>Default aggregation coordinator</dfn> is an [=aggregation coordinator=]
 that controls which is used for a report if none is explicitly selected.
 
-<dfn>Maximum report contributions</dfn> is a positive integer that controls how
-many contributions can be present in a single report.
+<dfn>Maximum report contributions</dfn> is a [=map=] from [=context type=] to
+positive {{unsigned long}} values. Semantically, it defines the maximum number
```
alexmturner, line 455: Let's just say "positive integer" -- specs don't usually use WebIDL types except in actual interfaces.

spec.bs
```diff
@@ -451,8 +451,10 @@ controls which [=origins=] are valid [=aggregation coordinators=]. Every
...
+of contributions that can be present a single report for every kind of calling
+context, e.g.  Shared Storage.
```
alexmturner, line 457: nit: two spaces

spec.bs
```diff
@@ -451,8 +451,10 @@ controls which [=origins=] are valid [=aggregation coordinators=]. Every
 <dfn>Default aggregation coordinator</dfn> is an [=aggregation coordinator=]
 that controls which is used for a report if none is explicitly selected.
 
-<dfn>Maximum report contributions</dfn> is a positive integer that controls how
-many contributions can be present in a single report.
+<dfn>Maximum report contributions</dfn> is a [=map=] from [=context type=] to
```
alexmturner, line 454: I'm a little concerned about the layering here -- in theory, this spec is meant to be as agnostic as possible about the specs using PAA. So, ideally those specs would be the ones defining this constant (and we'd monkeypatch for now), but I'm not sure if there's a good way to do that.

alexmturner (commented):
Generally LG! One layering question that might need some discussion

spec.bs
dmcardle: Yeah, how do we express this dependency? Basically, we want each calling API to parameterize this spec with a value for *maximum report contributions*?

alexmturner: Thinking aloud -- we could maybe change the context type section to say that it's a _struct_ representing the kind of global scope, it contains a name (string) and a default maximum number of contributions (positive integer). (With some comments about the constraints on those values.)

spec.bs
```diff
@@ -451,8 +451,10 @@ controls which [=origins=] are valid [=aggregation coordinators=]. Every
...
+positive integers. Semantically, it defines the maximum number of contributions
+that can be present a single report for every kind of calling context, e.g.
```
alexmturner, line 456: nit: "present in a single report"

alexmturner (approved):
LGTM w one more nit

PR #151, opened 2024-07-29
Spec: disallow fenced frames setting contextId/filteringIdMaxBytes.

blu25:
[`sharedStorage.run()`](https://wicg.github.io/shared-storage/#dom-sharedstorageworklet-run) accepts a `privateAggregationConfig`(https://patcg-individual-drafts.github.io/private-aggregation-api/#dom-sharedstoragerunoperationmethodoptions-privateaggregationconfig) field, which allows
settings to be passed in to be used when sending Private Aggregation reports. However, this can be used from within a fenced frame to exfiltrate information. Because of that, these fields must be completely disabled inside of fenced frames.

Note: The `aggregationCoordinatorOrigin` field is needed for to select the coordinator, which is a major part of the functionality of the API. The long-term goal is to disallow setting `privateAggregationConfig` altogether inside of fenced frames. But for now, the usability hit is way too large to disable all of the fields. This will be revisited in the future when there are alternate ways to set the coordinator origin.

***
<a href="https://pr-preview.s3.amazonaws.com/blu25/private-aggregation-api/pull/151.html" title="Last updated on Aug 22, 2024, 5:26 PM UTC (b8a8ff4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/151/80bc584...blu25:b8a8ff4.html" title="Last updated on Aug 22, 2024, 5:26 PM UTC (b8a8ff4)">Diff</a>


spec.bs
```diff
@@ -1145,38 +1148,45 @@ steps. They return an [=aggregation coordinator=], null or a {{DOMException}}:
...
-modified in four ways. First, add the following steps just after step 2 ("If
-{{Worklet/addModule()}} has not yet been called, ..."), renumbering later steps
-as appropriate:
+The {{SharedStorageWorklet}}'s {{SharedStorageWorklet/run()}} method steps are
+modified in four ways. First, add the following steps just after step 4 ("[=Assert=]: window is a
+[=Window=]"), renumbering later steps as appropriate:
 <div algorithm="shared-storage-run-monkey-patch-1">
-3. Let |preSpecifiedParams| be the result of [=obtaining the pre-specified
-    report parameters=] given |options|.
+5. Let |context| be <var ignore>window</var>'s [=Window/browsing context=].
```
alexmturner, line 1186: This part does seem to be less about our integration with shared storage than the shared storage spec itself, but probably it's fine to just monkey patch for now. (I'm hoping to move this all into the shared storage spec soon.)

spec.bs
```diff
@@ -40,6 +40,9 @@ spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
     type: dfn
         text: getting a property; url: dfn-getting-properties
         text: error code; url: dfn-error-code
+spec: fenced-frame; urlPrefix: https://wicg.github.io/fenced-frame/
+    type: dfn;
+        for: browsing context; text: fenced frame config instance
```
alexmturner, line 45: This is not actually linking to the right thing automatically. We might need to add `url: #browsing-context-fenced-frame-config-instance`

spec.bs
```diff
@@ -1145,38 +1148,45 @@ steps. They return an [=aggregation coordinator=], null or a {{DOMException}}:
 1. Return |origin|.
 
 To <dfn algorithm>obtain the pre-specified report parameters</dfn> given a
-{{SharedStorageRunOperationMethodOptions}} |options|, perform the following
-steps. They return a [=pre-specified report parameters=], null, or a
+{{SharedStorageRunOperationMethodOptions}} |options| and a [=Window/browsing context=] |context|,
```
alexmturner, line 1151: We might want to link just to `[=browsing context=]` (i.e. the type, not the field on window)

spec.bs
```diff
@@ -1145,38 +1148,45 @@ steps. They return an [=aggregation coordinator=], null or a {{DOMException}}:
...
     {{DOMException}} with name "`DataError`".
+1. If |calledFromFencedFrame| is true and |contextId| is not null, return a new {{DOMException}}
+    with name "`DataError`".
 1. Let |filteringIdMaxBytes| be the [=default filtering ID max bytes=].
 1. If |privateAggregationConfig|["{{SharedStoragePrivateAggregationConfig/filteringIdMaxBytes}}"]
     [=map/exists=], set |filteringIdMaxBytes| to
     |privateAggregationConfig|["{{SharedStoragePrivateAggregationConfig/filteringIdMaxBytes}}"].
 1. If |filteringIdMaxBytes| is not [=set/contained=] in the [=valid filtering ID
     max bytes range=], return a new {{DOMException}} with name "`DataError`".
+1. If |calledFromFencedFrame| is true and |filteringIdMaxBytes| is not the [=default filtering ID
```
alexmturner, line 1174: nit: Maybe we could consolidate these into something like:
```
1. If |context|'s ... instance=] is not null:
  1. If |contextId| is not null or ..., return a new {{DOMException}} ...
```

Might slightly help readability if we put them together, wdyt? We could also consider adding a note explaining the why.

spec.bs
```diff
@@ -1234,12 +1244,11 @@ Finally, at the end of the same nested scope, add the following step:
 </div>
 
 The {{WindowSharedStorage}}'s {{WindowSharedStorage/selectURL()}} method steps
-are modified in three ways. First, add the following steps just after step 5
-("If {{Worklet/addModule()}} has not yet been called, ..."), renumbering later
-steps:
+are modified in three ways. First, add the following steps just after step 6 ("If context is null
```
alexmturner, line 1247: nit: `"If <var ignore>context</var> is null ..."`

spec.bs
```diff
@@ -1145,38 +1148,45 @@ steps. They return an [=aggregation coordinator=], null or a {{DOMException}}:
...
     :: |contextId|
     : [=pre-specified report parameters/filtering ID max bytes=]
     :: |filteringIdMaxBytes|
 
-The {{WindowSharedStorage}}'s {{WindowSharedStorage/run()}} method steps are
-modified in four ways. First, add the following steps just after step 2 ("If
-{{Worklet/addModule()}} has not yet been called, ..."), renumbering later steps
-as appropriate:
+The {{SharedStorageWorklet}}'s {{SharedStorageWorklet/run()}} method steps are
+modified in four ways. First, add the following steps just after step 4 ("[=Assert=]: window is a
```
alexmturner, line 1183: nit: `"[=Assert=]: <var ignore>window</var> is a...`

alexmturner (commented):
Thanks, generally looks good! Just some nits

spec.bs
blu25: Added the url to the entry.

spec.bs
blu25: Changed it to the browsing context type.

spec.bs
blu25: Added `<var ignore>`.

spec.bs
blu25: Added `<var ignore>`.

spec.bs
blu25: Ack. Will leave as is for now.

spec.bs
blu25: Consolidated into one section in a nested if statement.

PR #152, opened 2024-07-31
Spec: Fix minor type issue in consume budget if permitted

alexmturner:
The algorithm takes a long when it should just take an integer (as the sum of many longs may not fit in a long).

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/152.html" title="Last updated on Aug 1, 2024, 6:25 PM UTC (d844e5b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/152/b87549a...d844e5b.html" title="Last updated on Aug 1, 2024, 6:25 PM UTC (d844e5b)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -697,8 +697,8 @@ null |timeout|:
     |currentWallTime|.
 1. [=set/Append=] |report| to the user agent's [=aggregatable report cache=].
 
-To <dfn algorithm>consume budget if permitted</dfn> given a {{long}} |value|, an
-[=origin=] <var ignore=''>origin</var>, a [=context type=] |api| and a
+To <dfn algorithm>consume budget if permitted</dfn> given an integer |value|,
+an [=origin=] <var ignore=''>origin</var>, a [=context type=] |api| and a
 [=moment=] |currentTime|, perform [=implementation-defined=] steps. They return
 a [=boolean=], which indicates whether there is sufficient 'contribution budget'
 left to send the requested contribution |value|. This budget should be bound to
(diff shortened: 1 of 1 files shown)


alexmturner:
@dmcardle, could you PTAL? Thanks!

spec.bs
```diff
@@ -697,8 +697,8 @@ null |timeout|:
     |currentWallTime|.
 1. [=set/Append=] |report| to the user agent's [=aggregatable report cache=].
 
-To <dfn algorithm>consume budget if permitted</dfn> given a {{long}} |value|, an
-[=origin=] <var ignore=''>origin</var>, a [=context type=] |api| and a
+To <dfn algorithm>consume budget if permitted</dfn> given an integer |value|,
```
dmcardle, line 700: Maybe also specify that it's the sum of contribution values?

I also just realized that we never actually "call" this algorithm anywhere. Is that intentional?

alexmturner: Updated the wording slightly, lmk if that's clearer. And we do call it [here](https://patcg-individual-drafts.github.io/private-aggregation-api/#ref-for-consume-budget-if-permitted%E2%91%A0) :)

dmcardle: Aha! I couldn't find the reference because I was searching for `[=consume` in spec.bs and found nothing. Instead, we're calling  "consum**ing** budget if permitted" [here](https://github.com/patcg-individual-drafts/private-aggregation-api/blob/b87549a374f24616c467bf77e7f786b4a9a8c212/spec.bs#L682). Despite not being a literal string match, bikeshed correctly links to the definition, so maybe it's doing some fuzzy matching?

dmcardle (approved):
LGTM!

spec.bs
alexmturner: Yep, bikeshed is very cool! It'll match like this for the first or last word in the name (see [here](https://speced.github.io/bikeshed/#conjugating)). For future reference, you can click on the definition in the rendered file and it shows all of the references.

alexmturner:
Thanks for the review!

PR #155, opened 2024-08-19
Spec: Improve definition of the encryption algorithm

dmcardle:
This adds a missing return step to the algorithm definition, adds an explicit instruction to throw when encryption fails, and attempts to clarify our commentary on RFC9180.

Followup to #147

***
<a href="https://pr-preview.s3.amazonaws.com/dmcardle/private-aggregation-api/pull/155.html" title="Last updated on Aug 22, 2024, 2:52 PM UTC (744269b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/155/49cd4e0...dmcardle:744269b.html" title="Last updated on Aug 22, 2024, 2:52 PM UTC (744269b)">Diff</a>


dmcardle:
@alexmturner, PTAL!

Direct preview link to the modified algorithm: https://pr-preview.s3.amazonaws.com/dmcardle/private-aggregation-api/pull/155.html#encrypt-the-payload

spec.bs
```diff
@@ -1060,22 +1060,28 @@ To <dfn>obtain the plaintext payload</dfn> given an [=aggregatable report=]
 <div algorithm>
 To <dfn>encrypt the payload</dfn> given a [=byte sequence=] |plaintextPayload|,
 public key |pkR| and a [=string=] |sharedInfo|, perform the following steps.
-They return a [=byte sequence=] or an error.
+They either return a [=byte sequence=] or throw an error.
 
 1. Let |info| be the result of [=UTF-8 encoding=] the [=string/concatenation=]
```
dmcardle, line 1065: I'm tempted to assert here that the concatenation will contain only ASCII characters.

spec.bs
```diff
@@ -1060,22 +1060,28 @@ To <dfn>obtain the plaintext payload</dfn> given an [=aggregatable report=]
...
+    Note: The |ciphersuite| triple is composed of HPKE
+        [[RFC9180#name-algorithm-identifiers|algorithm identifiers]], specifying
+        the [[RFC9180#name-key-encapsulation-mechanism|KEM]] as DHKEM(X25519,
+        HKDF-SHA256), the [[RFC9180#name-key-derivation-functions-kd|KDF]]
+        function as HKDF-SHA256 and the
+        [[RFC9180#name-authenticated-encryption-wi|AEAD]] function as
+        ChaCha20Poly1305.
+1. Let |encryptedPayload| be the result of calling HPKE's
+    [[RFC9180#single-shot-apis|single-shot encryption function]]: <var
+    ignore>Seal&lt;Base&gt;</var>(|pkR|, |info|, |aad|, |plaintextPayload|).
```
alexmturner, line 1079: nit: I think the RFC expects `<MODE>` to be replaced with `Base` (i.e. losing the angle brackets).

But also I'm wondering if it might be clearer to keep the more prose-y description from before given that the functions defined in the RFC have important parameters omitted (no idea why they do that...).

alexmturner (commented):
Thanks! Couple comments

spec.bs
dmcardle: OK, I added an assertion step. I'll leave this open so we remember to at least file a crbug for our implementation.

dmcardle: You know, I think we can drop the assertion. While I believe the sharedInfo string is guaranteed to be ASCII, the spec is about to pass it into a UTF-8 encoder, so it doesn't really matter.

alexmturner: Ack

spec.bs
alexmturner: Definitely happy with changes to make those two points you enumerated clearer. For the second point, we should maybe be explicitly saying something like:
```
1. Let |encryptedPayload| be the concatenation of |enc| and |ct|.
```
One caveat is that the infra spec doesn't actually define concatenation for byte sequences, but I think it's pretty clear anyway. Also, given we need |enc| and |ct| separately, the benefit of using the single-shot API might be lessened. (Mainly I'm just wondering if the <MODE> parameter is confusing enough to make the separate calls better.)

Also don't feel too strongly on prose vs non-prose, but maybe we can split the difference and call it in the web spec algorithm style, but referring to the exact spec pseudocode names, e.g.:
```
1. Let (|enc|, |hpkeContext|) be the result of setting up an
  [[RFC9180|HPKE]][[RFC9180#name-encryption-to-a-public-key|sender's context]] by calling
  SetupBaseS() with a public key |pkR|, application-supplied information |info| and ciphersuite
  |ciphersuite|. If this operation fails, return an error.

  Note: Although RFC9180 omits the parameter from its pseudocode, |ciphersuite| is passed
    implicitly.
```
Wdyt?

spec.bs
```diff
@@ -1062,20 +1062,27 @@ To <dfn>encrypt the payload</dfn> given a [=byte sequence=] |plaintextPayload|,
...
+    Although RFC9180 omits the parameter from its pseudocode, |ciphersuite| is
+    passed implicitly.  All of RFC9180's high-level HPKE operations
+    [[RFC9180#name-errors|are fallible]]; if encryption fails, throw
+    [=exception/throw=] a {{TypeError}}.
```
alexmturner, line 1084: nit: return the error

spec.bs
```diff
@@ -1062,20 +1062,27 @@ To <dfn>encrypt the payload</dfn> given a [=byte sequence=] |plaintextPayload|,
...
-        function as DHKEM(X25519, HKDF-SHA256), the KDF function as HKDF-SHA256
-        and the AEAD function as ChaCha20Poly1305.
-1. Let |hpkeContext| be the result of setting up an [[RFC9180|HPKE]]
-    [[RFC9180#name-encryption-to-a-public-key|sender's context]] with |pkR|,
-    |info|, |kem_id|, |kdf_id| and |aead_id|.
 1. Let |aad| be \`\` (an empty [=byte sequence=]).
-1. Let <var ignore>encryptedPayload</var> be the result of
-    [[RFC9180#name-encryption-and-decryption|encrypting]] |plaintextPayload|
-    with |hpkeContext| and |aad|.
+1. Let |ciphersuite| be (0x0020, 0x0001, 0x0003).
```
alexmturner, line 1069: very optional: the names kem_id, kdf_id and aead_id might align slightly better with the HPKE spec, but either way let's definitely keep the links to each section in the Note.

spec.bs
dmcardle: > One caveat is that the infra spec doesn't actually define concatenation for byte sequences, but I think it's pretty clear anyway.

Yeah, agreed. Not sure it's worth pointing out in the spec because I'm sure an implementer can figure out how to concatenate byte sequences.

> Also, given we need |enc| and |ct| separately, the benefit of using the single-shot API might be lessened. (Mainly I'm just wondering if the parameter is confusing enough to make the separate calls better.)

The single-shot encryption function returns the pair (`enc`, `ct`) and hides the HPKE context object. I think it's slightly simpler, but it does make for a kind of bloated step. I don't feel too strongly.

spec.bs
dmcardle: Fair point, let's do that.

dmcardle:
@alexmturner I took your advice and made a few edits, PTAL at this latest revision!

spec.bs
```diff
@@ -1060,27 +1060,40 @@ To <dfn>obtain the plaintext payload</dfn> given an [=aggregatable report=]
...
-1. Let |encryptedPayload| be the result of calling HPKE's
-    [[RFC9180#single-shot-apis|single-shot encryption function]]: <var
-    ignore>Seal&lt;Base&gt;</var>(|pkR|, |info|, |aad|, |plaintextPayload|).
-    Although RFC9180 omits the parameter from its pseudocode, |ciphersuite| is
-    passed implicitly.  All of RFC9180's high-level HPKE operations
-    [[RFC9180#name-errors|are fallible]]; if encryption fails, throw
-    [=exception/throw=] a {{TypeError}}.
+1. Let (|enc|, |hpkeContext|) be the result of setting up an [[RFC9180|HPKE]]
+    [[RFC9180#name-encryption-to-a-public-key|sender's context]] by calling <var
+    ignore>SetupBaseS()</var> with a public key |pkR|, application-supplied
```
alexmturner, line 1078: nit: Should SetupBaseS() be in backticks (i.e. monospace font) instead of treating as a variable (italics)?

ditto for Seal() below (and the other SetupBaseS())

alexmturner (approved):
Thanks, LGTM w one nit! (and needing a rebase after #156)

spec.bs
dmcardle: Sure, that works. I have no idea what the convention is for referencing familiar categories of objects that are defined outside of the Bikeshed bubble.

The mixing and matching is still a little weird, but I think it's unavoidable. Rather than saying "call *hpkeContext.Seal()* with ...", now I'll say "call `ContextS.Seal()` on the *hpkeContext* object with ...."

spec.bs
```diff
@@ -1066,16 +1066,35 @@ They return a [=byte sequence=] or an error.
...
-    with |hpkeContext| and |aad|.
+1. Let |ciphertext| be the result of
+    [[RFC9180#name-encryption-and-decryption|sealing]] the payload by calling
+    `ContextS.Seal()` on the |hpkeContext| object with additional authenticated
+    data |aad| and plaintext |plaintextPayload|. If this operation fails, return
+    an error.
+1. Let |encryptedPayload| be the concatenation of the [=byte sequences=] «
+    |enc|, |ciphertext| ».
+
+    Note: The length of the encapsulated symmetric key <var ignore>enc</var>
```
dmcardle, line 1094: Pedantically, I changed `enc` to a `<var ignore>` because this reference should not increase its usage count.

alexmturner (approved):
LGTM, thanks!

PR #156, opened 2024-08-21
Spec: Fix broken references to Shared Storage

dmcardle:
Looks like a change upstream in the Shared Storage spec means we need to update our references: https://github.com/WICG/shared-storage/pull/171

The "Go through all monkey patches" issue on line 1166 still stands :)

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1243,8 +1243,8 @@ steps. They return a [=pre-specified report parameters=], null, or a
 
 </div>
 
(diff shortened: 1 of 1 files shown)


dmcardle:
I just realized "The SharedStorage's run() method" doesn't flow as well as "The SharedStorageWindow's run() method", so I dropped the possessive.

alexmturner (approved):

PR #158, opened 2024-08-28
Add Makefile for local dev: building spec, rendering Markdown, etc.

dmcardle:
Although the spec is already compiled automatically by CI and GitHub renders our Markdown, I still find it useful to automate commands that I frequently use in local development. A Makefile is better than another Markdown file because it's executable!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ spec.html: spec.bs
 README-with-toc.md: README.md
 	pandoc -f gfm --toc --toc-depth 6 -s $< -o $@
 
-# Rule for generating HTML from Markdown for a limited set of targets. This is a
+# Rule for generating HTML from Markdown for a limited set of targets. This uses
 # GNU Make "static pattern" syntax.
 $(HTML_FROM_MD_TARGETS): %.html : %.md
 	pandoc -f gfm -s $< -o $@
--- a/Makefile
+++ b/Makefile
@@ -13,10 +13,10 @@ clean:
 # out to stale specs. https://speced.github.io/bikeshed/#cli-update
 .PHONY: update
 update:
-	python3 -m bikeshed update
+	bikeshed update
 
 spec.html: spec.bs
-	python3 -m bikeshed --die-on=everything spec $< $@
+	bikeshed --die-on=everything spec $< $@
 
 # Autogenerates a table of contents for the README. This can in turn be rendered
 # as HTML. This is useful for catching mistakes in the README's handwritten TOC.
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,29 @@
+# HTML files that are generated from Markdown sources.
+HTML_FROM_MD_TARGETS=README.html README-with-toc.html
+
+.PHONY: all
+all: spec.html $(HTML_FROM_MD_TARGETS)
+
+.PHONY: clean
+clean:
+	-rm spec.html
+	-rm $(HTML_FROM_MD_TARGETS)
+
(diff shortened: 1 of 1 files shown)


dmcardle:
@alexmturner WDYT? I've been keeping this under my hat for a while, but maybe it's worth putting in the repo.

dmcardle:
Ah, `python3 -m bikeshed` wasn't compatible with bikeshed's recommended pipx installation on my Linux machine. Pushed a fixup commit.

Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+.PHONY: all
+all: spec.html $(HTML_FROM_MD_TARGETS)
+
+.PHONY: clean
+clean:
+	-rm spec.html
+	-rm $(HTML_FROM_MD_TARGETS)
+
+# Updates Bikeshed's datafiles. Run this regularly to ensure you're not linking
```
alexmturner, line 12: This might not be necessary, bikeshed should update automatically ([ref](https://speced.github.io/bikeshed/#:~:text=By%20default%2C%20Bikeshed%20checks%20how,than%20a%20few%20days%20old)), so might want to update the comment

Makefile
```diff
@@ -0,0 +1,29 @@
...
+# out to stale specs. https://speced.github.io/bikeshed/#cli-update
+.PHONY: update
+update:
+	bikeshed update
+
+spec.html: spec.bs
+	bikeshed --die-on=everything spec $< $@
```
alexmturner, line 19: Do we want to die on everything as the default?

alexmturner (commented):
Generally lg, bit of a rubber stamp as I'm not super knowledgeable about makefiles -- but put a couple nits

Makefile
dmcardle: Oh weird, I actually added this because I hit a scenario where my local build succeeded but CI failed. I suspect it's due to the  "more than a few days old" part.

I can sharpen the comment a bit:

```
# Updates Bikeshed's datafiles. Bikeshed takes an eventual-consistency approach
# to updating its datafiles. Consequently, local builds may incorrectly succeed
# when referenced documents were recently updated.
```

WDYT?

Makefile
dmcardle: I like having a really strict local build and it hasn't caused issues so far. If it's annoying, we can always weaken this?

alexmturner: Personally, I find it easier to have all the errors listed at once, but I'm also happy to just run `bikeshed`, so we can keep this.

Makefile
alexmturner: Maybe let's keep a reference to the time frame, e.g. something like:
Updates Bikeshed's datafiles. Bikeshed automatically updates them if they're more than a few days old, but this will ensure you have the latest version.

dmcardle: That's a little clearer, thanks.

Makefile
dmcardle: Oh, actually `--die-on` just makes the warnings fatal. Bikeshed will still emit all errors because  `--die-when` defaults to `late` ([docs](https://speced.github.io/bikeshed/#:~:text=If%20unspecified%2C%20this%20defaults%20to%20late.)).

dmcardle:
Thanks! Ready to go, I think :)

Makefile
alexmturner: Empirically, before the last commit landed, I get one dfn error and three link errors when I call `bikeshed spec`, but no link errors when I call `bikeshed --die-on=everything spec`, but not certain why :)

alexmturner (approved):
LGTM, thanks

PR #163, opened 2024-10-14
Spec: export obtain the Private Aggregation coordinator algorithm

alexmturner:
The version that takes a USVString is shared between Protected Audience and Shared Storage, so we can export it. Updates links so that the monkey patched version has a modified linking text as that will be removed from this spec soon.

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/163.html" title="Last updated on Oct 17, 2024, 10:04 PM UTC (217a8e2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/163/f5cda2b...217a8e2.html" title="Last updated on Oct 17, 2024, 10:04 PM UTC (217a8e2)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -642,6 +642,24 @@ an [=origin=] |origin|, perform the following steps. They return a [=boolean=].
 </div>
 
 <div algorithm>
+To <dfn export>obtain the Private Aggregation coordinator</dfn> given a
+{{USVString}} |originString|, perform the following steps. They return an
+[=aggregation coordinator=] or a {{DOMException}}.
+
+1. Let |url| be the result of running the [=URL parser=] on |originString|.
+1. If |url| is failure or null, return a new {{DOMException}} with name
+    "`SyntaxError`".
+
+    Issue: Consider throwing an error if the path is not empty.
+1. Let |origin| be |url|'s [=url/origin=].
+1. If the result of [=determining if an origin is an aggregation coordinator=]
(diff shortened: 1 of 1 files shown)


qingxinwu (approved):
LGTM. Thanks

spec.bs
```diff
@@ -2119,37 +2138,18 @@ They return an [=interest group=] or null:
 </div>
 
 <div algorithm>
-To <dfn>obtain the Private Aggregation coordinator</dfn> given a
+To <dfn lt="obtain the Private Aggregation coordinator from a PA config">obtain
+the Private Aggregation coordinator</dfn> given a
```
qingxinwu, line 2142: nit: is there a reason to keep this the same name as the other? Is it slightly easier to just call it "... from a PA config"

alexmturner: I was treating them like "overloaded" functions, but I don't feel strongly. That being said, I don't love the "PA config" if it's the proper name -- maybe we could call it "obtain the coordinator from a Private Aggregation config"?

alexmturner: Went with this!

qingxinwu (approved):

spec.bs
qingxinwu: LGTM

qingxinwu (approved):

PR #171, opened 2025-01-23
Spec: Weaken null report requirement WRT max contributions

dmcardle:
It's not necessary to send a null report when the caller-provided "max contributions" has no effect on the payload length.

This commit updates the algorithm "determine if a report should be sent deterministically" to account for this edge case. Relatedly, it also updates the prose description of payload-length side channel.

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -572,14 +572,17 @@ scope=] |debugScope| and an optional [=debug details=] or null
 
 <div algorithm>
 To <dfn>determine if a report should be sent deterministically</dfn> given a
-[=pre-specified report parameters=] |preSpecifiedParams|, perform the following
-steps. They return a [=boolean=]:
+[=pre-specified report parameters=] |preSpecifiedParams| and a [=context type=]
+|api|, perform the following steps. They return a [=boolean=]:
 1. If |preSpecifiedParams|' [=pre-specified report parameters/context ID=] is
     not null, return true.
 1. If |preSpecifiedParams|' [=pre-specified report parameters/filtering ID max
     bytes=] is not the [=default filtering ID max bytes=], return true.
-1. If |preSpecifiedParams|' [=pre-specified report parameters/max
-    contributions=] is not null, return true.
+1. Let |effectiveMaxContributions| be the result of [=determining the max
+    contributions=] with |api| and |preSpecifiedParams|' [=pre-specified report
+    parameters/max contributions=].
+1. Let |defaultMaxContributions| be [=default maxContributions by API=][|api|].
+1. If |effectiveMaxContributions| is not |defaultMaxContributions|, return true.
 1. Return false.
 
 Note: It is sometimes necessary to send a 'null report' to conceal the fact that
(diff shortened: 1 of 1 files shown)


dmcardle:
@alexmturner PTAL!

dmcardle:
Ping @alexmturner :)

spec.bs
```diff
@@ -1297,17 +1301,25 @@ by a randomized amount of time to make it difficult to determine whether a
...
-namely how many contributions are included. To protect against this, the payload
-is padded to a fixed number of contributions.
+The length of the encrypted payload could additionally expose some cross-site
+information, namely the number of contributions present in the plaintext
+payload. To eliminate this side channel, Private Aggregation ensures that
+payloads contain a predetermined number of contributions prior to encryption,
+potentially truncating or padding with null contributions to match the target.
+
+The target number of contributions is based on [=pre-specified report
+parameters/max contributions=] when it is non-null — it is the caller's
```
alexmturner, line 1319: I'm not sure I understand the "it is the caller's ... cross-site data" clause. Maybe can just drop it?

spec.bs
```diff
@@ -572,14 +572,18 @@ scope=] |debugScope| and an optional [=debug details=] or null
 
 <div algorithm>
 To <dfn>determine if a report should be sent deterministically</dfn> given a
```
alexmturner, line 574: I think we need to update the calls to this algorithm as well

spec.bs
```diff
@@ -572,14 +572,18 @@ scope=] |debugScope| and an optional [=debug details=] or null
...
 1. If |preSpecifiedParams|' [=pre-specified report parameters/context ID=] is
     not null, return true.
 1. If |preSpecifiedParams|' [=pre-specified report parameters/filtering ID max
     bytes=] is not the [=default filtering ID max bytes=], return true.
-1. If |preSpecifiedParams|' [=pre-specified report parameters/max
-    contributions=] is not null, return true.
+1. Let |effectiveMaxContributions| be the result of [=determining the max
+    contributions=] with |api| and |preSpecifiedParams|' [=pre-specified report
+    parameters/max contributions=].
+1. Let |defaultMaxContributions| be the result of [=determining the max
```
alexmturner, line 584: This definitely works, but wondering if `[=default maxContributions by API=][|api|]` might be slightly more obvious

alexmturner (commented):
Thanks for pinging!

spec.bs
dmcardle: Ah, that was a little circuitous!

spec.bs
dmcardle: Done.

spec.bs
dmcardle: Ah, I was trying to call out that Private Aggregation assumes that `pre-specified report parameters/max contributions` is not a function of cross-site data. Maybe that's obvious? It certainly interrupts the paragraph's flow, so I'll delete it.

dmcardle: Did a little more rewriting here, PTAL.

alexmturner: Ah ack -- yeah we could consider expressly saying somewhere that "pre-specified report parameters" generally must be set from a first-party context. But I think it's ok as is too.

PR #174, opened 2025-02-26
Spec: don't export / define "remote end steps"

alexmturner:
We should instead link to the original webdriver definition as used in the webdriver spec. Also adds the arguments to the definition.

***
<a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/pull/174.html" title="Last updated on Feb 28, 2025, 5:12 PM UTC (ed744ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/patcg-individual-drafts/private-aggregation-api/174/d612b5c...ed744ae.html" title="Last updated on Feb 28, 2025, 5:12 PM UTC (ed744ae)">Diff</a>

### DIFF
--- a/spec.bs
+++ b/spec.bs
@@ -1236,8 +1236,9 @@ Set local testing mode {#set-local-testing-mode}
     </table>
 </figure>
 
-<div algorithm>
-The <dfn export>remote end steps</dfn> are:
+<div algorithm="remote end steps">
+The <a spec="webdriver2">remote end steps</a> given <var ignore>session</var>,
+[=URL=] <var ignore>variables</var> and |parameters| are:
 
 1. If |parameters| is not a JSON-formatted [[ECMASCRIPT#sec-objects|Object]],
     return a [=error|WebDriver error=] with [=error code=] [=invalid argument=].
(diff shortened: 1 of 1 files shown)


spec.bs
```diff
@@ -1236,8 +1236,9 @@ Set local testing mode {#set-local-testing-mode}
     </table>
 </figure>
 
-<div algorithm>
-The <dfn export>remote end steps</dfn> are:
+<div algorithm="remote end steps">
+The <a spec="webdriver2">remote end steps</a> given <var ignore>session</var>,
+[=URL=] <var ignore>variables</var> and |parameters| are:
```
qingxinwu, line 1241: is it needed to link to URL, or simply <var ignore>URL variables</var>? It seems "URL variables" is the argument's name, e.g., https://w3c.github.io/webdriver/#find-elements-from-element

alexmturner: Ah, good catch, thanks! I was confused by the space, but you're right.

alexmturner:
@dmcardle low priority, but could you PTAL? Thanks!

dmcardle (approved):
LGTM!

PR #176, opened 2025-03-04
Add recipe to Makefile that generates an "exports" report

dmcardle:
@alexmturner PTAL!


Makefile
```diff
@@ -8,6 +10,8 @@ all: spec.html $(HTML_FROM_MD_TARGETS)
 clean:
 	-rm spec.html
 	-rm $(HTML_FROM_MD_TARGETS)
+	-rm exports-report.txt
```
alexmturner, line 13: Do we need this one?

alexmturner (approved):
LGTM w nit!

Makefile
dmcardle: Not anymore! Good catch.

PR #178, opened 2025-03-07
Adjust pending report limit error handling

alexmturner:
Previously, the description implies that this error would only be triggered when there is no space for the report. To avoid such reports always being dropped (or needing to add special case exemptions), we modify the trigger to occur when the limit is just reached.

### DIFF
--- a/error_reporting.md
+++ b/error_reporting.md
@@ -114,8 +114,9 @@ Audience contexts:
   contributions were dropped due to the per-report limit
 - `reserved.empty-report-dropped`: a report was not scheduled as it had no
   contributions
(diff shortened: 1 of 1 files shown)


error_reporting.md
```diff
@@ -164,6 +166,22 @@ due to the budget limit. This “named budget” functionality also supports oth
 use cases and has been proposed in [a separate
 explainer](https://github.com/patcg-individual-drafts/private-aggregation-api/blob/main/named_budgets.md).
 
+### Two phase processing
+
+We propose a two phase processing model where first contributions that are not
+conditional on error events are tenatively processed. (For example, testing
```
alois-bissuel, line 172: Typo

Issue #141, opened 2024-07-09
More precisely define the contribution budget and window in the spec


alexmturner:
The spec currently leaves the budget as fully implementation-defined (see [here](https://patcg-individual-drafts.github.io/private-aggregation-api/#consume-budget-if-permitted)). The explainer, however, describes Chrome's [implementation plan](https://github.com/patcg-individual-drafts/private-aggregation-api#implementation-plan).

While it may be difficult to align on exact details across implementations, we should probably define the privacy unit more precisely while leaving some key parameters as implementation-defined (e.g. the precise epsilon/budget and perhaps the timespan used in the privacy unit).

PR #179, opened 2025-03-28
Spec: Add named budgets

dmcardle:


dmcardle:
Hey @alexmturner, here's a rough draft of named budgets. PTAL!

I had to spec out the concept of budgeting scopes, but I think there's room for improvement. I added struct fields like the duration of time and the calling API to demonstrate that the max budget (the field we use) is scoped to those things. I'm curious what you think.

FYI, I've also done a few novel things (well, novel to this spec). I used the "advisement" class to make a warning and then used [MathML](https://developer.mozilla.org/en-US/docs/Web/MathML) tags to make the equation look nice.

dmcardle:
> I had to spec out the concept of budgeting scopes...

Aha, I guess I should mention issue #141.

spec.bs
```diff
@@ -385,6 +448,31 @@ report=] unless their [=aggregatable report/debug details=] differ.
...
+:: A [=duration=]
+: <dfn ignore>caller type</dfn>
+:: A [=context type=]
+
+</dl>
+
+Issue: Currently, this spec only references [=budgeting scope/max budget=]. We
+need to decide whether deleting the unused fields will improve clarity.
+
+<h3 dfn-type=dfn>Named budget ledger</h3>
```
dmcardle, line 471: I suppose "ledger" is inaccurate because it doesn't record individual "transactions". WDYT about calling it a *Named budget balance* or maybe just a *Named balance*?

spec.bs
```diff
@@ -556,6 +644,9 @@ A [=user agent=] holds an <dfn>aggregation coordinator map</dfn>, which is a
 A [=user agent=] holds a <dfn>pre-specified report parameters map</dfn>, which
 is a [=map=] from [=batching scopes=] to [=pre-specified report parameters=].
 
+A [=user agent=] holds a <dfn>named budget ledgers map</dfn>, which is a [=map=]
+from [=batching scopes=] to [=lists=] of [=named budget ledgers=].
+
```
dmcardle, line 649: I realized I omitted a key part, which is that budget consumption against named budgets is persisted on disk and that implementations will set practical limits, like "no more than 25 distinct named budgets over 24 hours".

It's not clear to me what the "lifetime" of the user agent is. Can we model storing to the disk just by updating these structures held by the user agent? If so, maybe I should add a new "persistent storage" object held by the user agent and explicitly store records of budget consumption per minute.

spec.bs
```diff
@@ -131,13 +131,15 @@ interface PrivateAggregation {
...
                                          record<DOMString, any> contribution);
+  undefined reserveNamedBudgets(record<DOMString, double> namedBudgets);
   undefined enableDebugMode(optional PADebugModeOptions options = {});
 };
 
 dictionary PAHistogramContribution {
   required bigint bucket;
   required long value;
   bigint filteringId = 0;
+  DOMString budgetName;
```
dmcardle, line 142: Oops, I think we landed on `namedBudget` but I forgot to update this. TBD.

spec.bs
```diff
@@ -614,6 +705,55 @@ minimum delay to deliver an [=aggregatable report=].
 random delay to deliver an [=aggregatable report=]. This delay is additional to
 the [=minimum report delay=].
 
+<dfn>All budgeting scopes</dfn> is a non-empty [=list=] of [=budgeting scopes=].
+
+<div class="advisement"><span class="marker">WARNING:</span> When defining a
+[=budgeting scope=], be aware that certain values of [=budgeting scope/max
+budget=] are incompatible with the way that
+{{PrivateAggregation/reserveNamedBudgets}} interprets budget fractions.
+Implementers must verify that every slice <math><mi>n</mi></math> of their
```
alexmturner, line 714: Wondering if we just declare that only powers of two are ok? E.g. along the lines of "Implementers must use a max budget of the form 2^k with k <= 52. Note: This ensures that i/|maxBudget| can be exactly represented with a `Number` for all integral i in [0,|maxBudget|]."

Alternatively, I think we could probably support any value by incrementing the fraction up to the next valid double (i.e. by adding Math.floor(Math.log2(fraction)) * Number.EPSILON), but that's probably too hacky to be worth exploring if we can just require powers of two.

(Side note: do you know why pr-preview doesn't seem to be working? Would be nice to see all the rendered equations!)

alexmturner:
Nice! Did a first round of review :)

alexmturner (commented):
Helps if I actually flush my comments:

spec.bs
dmcardle: I think I might want to generalize even further. Sometimes I want these maps from strings to int to track budget remaining, but sometimes it's useful to track budget consumed. I think it might reduce the cognitive load of reading some of the new algorithm additions if we're able to say things like `[=ledger/Subtract=] |budgetConsumed| from |budgetRemaining|`. Anyway, that's still TBD.

spec.bs
dmcardle: Gotcha. Still mulling on this, but in the meantime I left an issue about moving the data away from the user agent and onto the {{PrivateAggregation}} object. But yeah, we might be able to avoid the field entirely?

dmcardle:
Hey @alexmturner, new "patchset" (fixup commit). PTAL when you get a chance!

spec.bs
```diff
@@ -230,6 +240,59 @@ Note: For forward compatibility, we do not throw an error for unrecognized
...
+            this step.
+    1. If |name|'s [=string/length=] is greater than [=maximum length of budget
+        name=], [=exception/throw=] a "{{DataError}}" {{DOMException}}.
+1. Let |ledgers| be a new [=list=] of [=named budget ledgers=].
+1. [=list/iterate|For each=] |budgetingScope| in [=all budgeting scopes=]:
+    1. Let |maxBudget| be |budgetingScope|'s [=budgeting scope/max budget=].
+    1. Let |remainingBudget| be |maxBudget|.
+    1. Let |ledger| be a new [=named budget ledger=].
+    1. [=map/For each=] |name| → |fraction| of |namedBudgets|:
+        1. If |fraction| is NaN, ±Inf, or -0, [=exception/throw=] a
```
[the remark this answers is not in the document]
alexmturner: Yeah, it not being closed under simple mathematical operations is a bit funny, but treating the values as "complete" doubles after the WebIDL validation seems reasonable to me.

spec.bs
```diff
@@ -230,6 +240,59 @@ Note: For forward compatibility, we do not throw an error for unrecognized
...
+            "{{DataError}}" {{DOMException}}.
+        1. If |fraction| is negative or greater than 1, [=exception/throw=] a
```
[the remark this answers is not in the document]
alexmturner: Ah oops yep I meant positive, not non-negative! I think I'd slightly lean towards disallowing zero as using the budgeting layer to ignore certain contributions seems a bit janky, but I do see your point that there might be potential uses.

spec.bs
```diff
@@ -614,6 +705,55 @@ minimum delay to deliver an [=aggregatable report=].
...
+        <mo>≤</mo>
+        <mi>32</mi>
+        <mo>}</mo>
+    </mrow>
+</math>, but other values are not guaranteed to be safe. For instance, a budget
+slice of 64 cannot be represented by a double fraction relative to a [=budgeting
+scope/max budget=] of 103.
+</div>
+
+<dfn>Maximum number of named budgets</dfn> is a positive integer that defines an
```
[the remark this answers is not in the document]
alexmturner: My bad! I think I didn't expand the lines above properly. I feel a bit on the fence about whether these should be implementation-defined, but defer to you.

spec.bs
```diff
@@ -169,7 +171,12 @@ Each {{PrivateAggregation}} object has the following fields:
         just certain ones. A return value of true indicates that this call will
         be processed using the default implementation in this spec. A return
         value that is an [=exception=] allows the embedding API's processing to
-        [=exception/throw=].
+        [=exception/throw=].  See [[#overriding-contributetohistogramonevent]]
+        for more details.
+: <dfn export>embedding API supports named budgets</dfn> (default false)
+:: A [=boolean=]. Embedding APIs that set this to true should define a
+    "<code>namedBudget</code>" field in their custom contribution dictionary
+    that accepts nullable {{DOMString}} values with a default of null.
```
dmcardle, line 179: @alexmturner Thoughts on this new field attached to {{PrivateAggregation}}?

I'm thinking that if embedding APIs use something other than {{PAHistogramContribution}} and also want to support named budgets...

* Their dictionary must contain a field for the budget name. I can't imagine they'd need something other than `DOMString? namedBudget = null;`.
* If they override default processing of `contributeToHistogramOnEvent(event, contribution)` when `contribution != null`, it's their responsibility to run our [=validate a contribution's named budget=] algorithm from their hook algorithm.

I've addressed the `namedBudget` field issue in a kind of strange way. Maybe we should define a WebIDL "trait" dictionary:

```ts
dictionary PAWithNamedBudget {
  DOMString? namedBudget = null;
};

// --- Embedding API's spec ---
dictionary MyCustomContribution : PAWithNamedBudget {
  Foo bucket;
  Bar value;
  Baz filteringId;
};
```

I tried to address the requirement to call our named budget validation algorithm down in the section titled 
 "Overriding contributeToHistogramOnEvent() processing".

spec.bs
```diff
@@ -319,30 +387,72 @@ Note: A later algorithm [=asserts=] that, for any [=contribution cache entry=]
...
-If the embedding API overrides processing for a call (i.e. the algorithm does
-not return true), it should  accept any `event` that is a [=string/
-concatenation=] of « "<code>reserved.</code>", |errorEvent| » where |errorEvent|
-is an [=internal error event=] and should accept any `contribution` with a
-[=converted to a JavaScript value|JavaScript value=] that is [=converted to an
-IDL value|convertible=] to the IDL type {{PAHistogramContribution}}. That is, it
-should not return an [=exception=] in those cases. However, the embedding API
-may accept additional `event`s or `contribution`s that would not be accepted by
-the default processing defined in this spec.
+<table class="complex data longlastcol">
```
dmcardle, line 405: The table's not strictly necessary, but I thought it's kind of nice for clarity.

spec.bs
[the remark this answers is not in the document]
dmcardle: Yeah, I guess that dueling implementations with different values would not play well unless we had some rectification logic (like don't reject overly-long budget names, try truncating and see if they're all still unique...). I'll move these to constants.

spec.bs
```diff
@@ -914,6 +1056,11 @@ parameters=] |preSpecifiedParams| and a [=moment=] or null |timeout|:
 1. Let |allUnmergedContributions| be the result of [=compiling all unmerged
     contributions=], given |reportingOrigin|, |api|, |pendingContributions|,
     |preSpecifiedParams|, |timeout| and |currentWallTime|.
+1. If [=named budget ledgers map=][|batchingScope|] [=map/exists=]:
```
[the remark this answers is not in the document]
dmcardle: OK, so I still have this in place, but "query the budget" synthesizes ledgers on demand when reserveNamedBudgets() was never called. I'm open to moving that logic or encapsulating it in a helper algorithm, like "get or create named budget ledgers".

spec.bs
alexmturner: Ack -- I think my one concern is just that the description "A named budget ledger tracks the remaining budget per named budget" doesn't quite address that the ledger entry starts unadjusted for previous usage, and that we also need to limit based on overall usage (not just per-name). But if we clarify that, then I think counting 'down' is fine. (Another naming option: named budget allocation?)

spec.bs
alexmturner: One other issue that comes to mind here is that for Protected Audience "batching scopes" are auctions, but the explainer says reservations are scoped to a single worklet/script runner context. Should we be using "debug scopes" here? 

One other complication with moving reservations onto the {{PrivateAggregation}} object is handling contributions triggered by fenced frames via `window.fence.reportEvent()` (as those contributions outlive the auction).

spec.bs
```diff
@@ -614,6 +756,19 @@ minimum delay to deliver an [=aggregatable report=].
...
+
+<div class="advisement"><span class="marker">WARNING:</span>
+Implementers must only use [=budgeting scope/max budgets=] of the form
+<math><msup><mn>2</mn><mn>k</mn></msup></math>, where
+<math><mn>0</mn><mo>≤</mo><mi>k</mi><mo>≤</mo><mn>52</mn></math>.  This ensures
+that all budget fractions
+<math><mi>n</mi><mo>/</mo><msup><mn>2</mn><mi>k</mi></msup></math> can be
+represented exactly by a {{double}}, for all integral <math><mi>n</mi></math> in
+the range
+<math><mo>[</mo><mn>0</mn><mo>,</mo><msup><mn>2</mn><mi>k</mi></msup><mo>]</mo></math>.
```
alexmturner, line 769: nit: consider using an infra spec range here

spec.bs
```diff
@@ -879,6 +1034,10 @@ cache entry=] or an [=exception=].
 1. If |contribution|["{{PAHistogramContribution/filteringId}}"] is not [=set/
     contained=] in [=the exclusive range|the range=] 0 to
     256<sup>|filteringIdMaxBytes|</sup>, exclusive, return a {{RangeError}}.
+
+1. Let |namedBudgetExists| be the result of [=validating a contribution's named
+    budget=] given |contribution| and |scopingDetails|.
```
alexmturner, line 1039: nit: |contribution| -> |contribution|["{{PAHistogramContribution/namedBudget}}"]

spec.bs
```diff
@@ -911,9 +1098,13 @@ parameters=] |preSpecifiedParams| and a [=moment=] or null |timeout|:
         <a href="https://github.com/privacysandbox/attestation">enrolled</a> or
         a limit on pending reports being reached.
 1. Let |currentWallTime| be the [=current wall time=].
+1. Let |ledgers| be a new [=list=].
+1. If [=named budget ledgers map=][|batchingScope|] [=map/exists=]:
+    1. Let |ledgers| be  [=named budget ledgers map=][|batchingScope|].
```
alexmturner, line 1103: nit: Set |ledgers| to

spec.bs
```diff
@@ -1068,34 +1260,79 @@ Note: Two {{PAHistogramContribution}}s for the same report can be merged if and
...
+[=list/size=] as |contributions|.
+
+1. If |remainingBudgetLedgers| is [=list/empty=]:
+    1. [=list/For each=] |budgetingScope| of [=all budgeting scopes=]:
+        1. Let |remainingBudgetLedger| be a new [=named budget ledger=].
+        1. [=map/Set=] |remainingBudgetLedger|[null] to |budgetingScope|'s
+            [=budgeting scope/max budget=].
+        1. [=list/Append=] |remainingBudgetLedger| to |remainingBudgetLedgers|.
+1. Let |numScopes| be the [=list/size=] of [=all budgeting scopes=].
+1. [=Assert=] that |numScopes| is equal to the [=list/size=] of
```
alexmturner, line 1276: Wondering if the ledgers should be a map or something if we go the route of actually enumerating the two 'scopes' (I.e. to avoid the parallel lists).

If we wanted to simplify this logic, we could also consider just computing allocations based off the 2^16 10-min limit and scaling it up 16x for the daily limit. We could also consider having the implementation-defined algorithm below just return a "budget left" for each name, taking into account both scopes. But it's up to you how much detail you want to put in this PR.

spec.bs
```diff
@@ -1068,34 +1260,79 @@ Note: Two {{PAHistogramContribution}}s for the same report can be merged if and
...
+    |remainingBudgetLedgers|.
+1. [=list/For each=] |i| in [=the exclusive range|the range=] from 0 to
+    |numScopes|, exclusive:
+    1. Let |budgetingScope| be [=all budgeting scopes=][|i|].
+    1. Let |duration| be |budgetingScope|'s [=budgeting scope/duration=].
+    1. Let |remainingBudgetLedger| be |remainingBudgetLedgers|[|i|].
+    1. [=map/For each=] |name| → |maxBudget| of |remainingBudgetLedger|:
+        1. Let |budgetConsumed| be an integer determined by an
```
alexmturner, line 1284: We also need to consider the case where |budgetConsumed| is small for a particular named budget, but usage of other names limits this one more than its current reservation. (This could happen if reservations aren't always consistent between calls.)

For this, I think we need an overall previous consumption or an overall budget left.

spec.bs
```diff
@@ -1068,34 +1260,79 @@ Note: Two {{PAHistogramContribution}}s for the same report can be merged if and
...
+
+    Note: At this point, we have adjusted the values of each ledger in
+    |remainingBudgetLedgers| based on |origin|'s historical budget consumption.
+    Next, we will accept contributions on a first-come, first-served basis, by
+    subtracting the value of each contribution from the ledgers in
+    |remainingBudgetLedgers| until one would become empty.
+
 1. Let |resultForEachContribution| be a new [=list=] of [=booleans=].
-1. Let |approvedValueSum| be 0.
+1. Let |ledgerConsumed| be a new [=named budget ledger=].
```
alexmturner, line 1306: This conflicts a bit with the named budget ledger definition above as "track[ing] the remaining budget". We might want to adjust the definition (or just maintain a map here).

spec.bs
```diff
@@ -180,9 +182,13 @@ The <dfn method for="PrivateAggregation">
 contributeToHistogram(PAHistogramContribution |contribution|)</dfn> method steps
 are:
 
+1. Let |scopingDetails| be [=this=]'s [=PrivateAggregation/scoping details=].
+1. Let |namedBudgetExists| be the result of [=checking that a contribution's
```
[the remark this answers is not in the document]
alexmturner: I was thinking we could potentially do the validation before calling the delegation function, but I don't think that would align with our likely implementation. So, yeah probably doesn't affect it.

spec.bs
```diff
@@ -614,6 +705,55 @@ minimum delay to deliver an [=aggregatable report=].
 random delay to deliver an [=aggregatable report=]. This delay is additional to
 the [=minimum report delay=].
 
+<dfn>All budgeting scopes</dfn> is a non-empty [=list=] of [=budgeting scopes=].
+
+<div class="advisement"><span class="marker">WARNING:</span> When defining a
```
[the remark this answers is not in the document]
alexmturner: oh weird -- yeah agreed that the latter seems better! I'm fine with the warning text

alexmturner (commented):
Thanks, Dan! Did a pass :)

